PR #14, opened 2018-03-23
Add access prohibition for external links

rugk:
Please verify yourself, whether this actually works. I've just modified this based on looking at the source.


tasn:
Thanks. Released and deployed.

PR #23, opened 2018-11-21
Disable referrer when clicking links

rugk:
I guess this won't hurt for privacy, even if there is no referrer on the page. :smile:


tasn:
Thanks!

PR #57, opened 2019-10-15
Fix display of recurring events.  Fixes #56

### LINKED ISSUE #56: "Event should be defined" when viewing recurring event
After a recurring event is created (i.e. via android stock calendar app) the detail of the first event is displayed correctly when clicking on the event.  For the other events the following error is created:
```
render
../etesync-web/src/components/Event.tsx:16
finishClassComponent
../etesync-web/node_modules/react-dom/cjs/react-dom.development.js:14979
updateClassComponent
../etesync-web/node_modules/react-dom/cjs/react-dom.development.js:14934
beginWork
../etesync-web/node_modules/react-dom/cjs/react-dom.development.js:15840
performUnitOfWork
../etesync-web/node_modules/react-dom/cjs/react-dom.development.js:19745
workLoop
../etesync-web/node_modules/react-dom/cjs/react-dom.development.js:19786
callCallback
../etesync-web/node_modules/react-dom/cjs/react-dom.development.js:147
invokeGuardedCallbackDev
../etesync-web/node_modules/react-dom/cjs/react-dom.development.js:196
invokeGuardedCallback
../etesync-web/node_modules/react-dom/cjs/react-dom.development.js:250
replayUnitOfWork
../etesync-web/node_modules/react-dom/cjs/react-dom.development.js:18976
renderRoot
../etesync-web/node_modules/react-dom/cjs/react-dom.development.js:19899
performWorkOnRoot
../etesync-web/node_modules/react-dom/cjs/react-dom.development.js:20811
performWork
../etesync-web/node_modules/react-dom/cjs/react-dom.development.js:20721
performSyncWork
../etesync-web/node_modules/react-dom/cjs/react-dom.development.js:20695
interactiveUpdates$1
../etesync-web/node_modules/react-dom/cjs/react-dom.development.js:20983
interactiveUpdates
../etesync-web/node_modules/react-dom/cjs/react-dom.development.js:2267
dispatchInteractiveEvent
../etesync-web/node_modules/react-dom/cjs/react-dom.development.js:5081
```
The cause is that the subsequent recurring react events components don't have the props.item set.
```
  public render() {
    if (this.props.item === undefined) {
      throw Error('Event should be defined!');
    }
```
With my limited knowledge it looks like the issue is in PimMain.tsx at this code snipped
```
  public eventClicked(event: ICAL.Event) {
    // FIXME:Hack
    const itemUid = `${(event as any).journalUid}|${event.uid}`;
```
where `${(event as any).journalUid}` returns "undefined".

This traces back to the render method of Calendar.tsx where the event is cloned. This clone does not replicate the `journalUid` property of the event.

Thinking of being smart I added the following statement after the clone
```
ev.journalUid = event.journalUid
```
which caused the error message: "Property 'journalUid' does not exist on type 'EventType'."
I think that explains also why the clone dropped the property.

etesync

cniesen:


tasn:
Oops, replied at the wrong place. Moved it here:

Hey,

Thanks for the patch! Just two comments:

1. Could you please use spaces and not tabs (the same indentation like the line before).
2. Could you please move the "Fixes etesync#56" part to the commit message itself and not the title? So:
```
Fix display of recurring events.

Fixes 56
```

Thanks!

cniesen:
OK.  I think now I got it right. :)

tasn:
Thanks a lot!

Issue #50, opened 2019-09-02
Add support for recurring events


neurokrish:
Can't seem to find a way to add recurring events. Will be very useful for events like birthdays..

tasn:
Definitely. FWIW they currently show correctly, you just can't edit them.

neurokrish:
Doesn't work for me. I have a birthday calendar with recurring yearly events imported. When I click those events, it throws up an error saying "something went wrong" and that an "event should be defined!" with something under "stack trace". No such errors occur for regular calendar events. Am I missing something?

neurokrish:
OK, I misread your comment. Looks like you know the issue. Yes, I can see the recurring events just fine but when I click on it, I get an error. Guess that's what you meant by "can't edit them". Hope you can push a fix for this.

tasn:
I'm actually not familiar with the crash. It should show a nice error message. Could you maybe put the backtrace here? Thanks!

Nuntius0:
```
Error: Event should be defined!
    at t.value (https://client.etesync.com/static/js/main.59c3899d.chunk.js:1:36954)
    at Mo (https://client.etesync.com/static/js/2.88819104.chunk.js:1:763666)
    at Co (https://client.etesync.com/static/js/2.88819104.chunk.js:1:763459)
    at Io (https://client.etesync.com/static/js/2.88819104.chunk.js:1:767316)
    at Ya (https://client.etesync.com/static/js/2.88819104.chunk.js:1:786901)
    at qa (https://client.etesync.com/static/js/2.88819104.chunk.js:1:787285)
    at Ts (https://client.etesync.com/static/js/2.88819104.chunk.js:1:793831)
    at ks (https://client.etesync.com/static/js/2.88819104.chunk.js:1:793211)
    at Ds (https://client.etesync.com/static/js/2.88819104.chunk.js:1:799862)
    at Tn (https://client.etesync.com/static/js/2.88819104.chunk.js:1:732363)
```
I'm getting this too. Weirdly enough, only for some recurring events but not all, and sometimes only when clicking on the event from the week view but not from the month view.

tasn:
This is a recent regression, I swear it used to work. I wonder what it's about.

tasn:
Btw, @Nuntius0, let's talk about this issue in #56. This ticket is about adding a way to add recurring events.

PR #69, opened 2019-12-19
widgets: add RRule widget

Tal-Leibman:
First step towards fixing  #50
RRule widget for repeating tasks and calendar events


src/widgets/RRule.tsx
```diff
@@ -0,0 +1,101 @@
+import * as React from 'react';
+import Container from './Container';
+import { Input, RadioGroup, FormControlLabel, Radio, FormLabel } from '@material-ui/core';
+import RRule, { RRuleSet, Frequency } from 'rrule';
+
+interface PropsType { sendRRule: (rRule: string) => void }
```
tasn: Rename:
sendRRule -> onChange
rRule -> rrule

And also make it multiline.

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+export default function RRuleEteSync(props: PropsType) {
+  const [frequency, setFrequency] = React.useState(RRuleSet.HOURLY);
+  const [interval, setInterval] = React.useState<number | undefined>();
+  const [ends, setEnds] = React.useState('never');
+  const [until, setUntil] = React.useState<Date | null>(null);
+  const [count, setCount] = React.useState<number | null | undefined>();
+  const rule = new RRule({
```
tasn: No need to actually create an rrule object here. You can just have an object that looks like what you're passing to the RRule constructor as the main rule object.

package.json
```diff
@@ -27,6 +27,7 @@
     "redux-persist": "^5.10.0",
     "redux-thunk": "^2.3.0",
     "reselect": "^3.0.1",
+    "rrule": "^2.6.3",
```
tasn: No need to import rrule, you can just make it "rrule like" I'll explain more below.

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,101 @@
...
+          type="date"
+          value={until?.toISOString().split('T')[0]}
+          onChange={(event: React.FormEvent<{ value: unknown }>) => {
+            const dateInput = event.currentTarget as HTMLInputElement;
+            setUntil(dateInput.valueAsDate);
+          }}
+        />
+      </Container>
+      <Container>
+        {rule.toText()}
```
tasn: Not needed.

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,101 @@
...
+            event.preventDefault();
+            const inputNode = event.currentTarget as HTMLInputElement;
+            if (inputNode.value === '') {
+              setCount(null);
+            } else if (inputNode.valueAsNumber) {
+              setCount(inputNode.valueAsNumber);
+            }
+          }}
+        />
+        <Input
```
tasn: As said elsewhere, use the same picker we use in the rest of the code (e.g. EventEdit.tsx)

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,101 @@
...
+            event.preventDefault();
+            const inputNode = event.currentTarget as HTMLInputElement;
+            if (inputNode.value === '') {
+              setInterval(undefined);
+            } else if (inputNode.valueAsNumber) {
+              setInterval(inputNode.valueAsNumber);
+            }
+          }}
+        />
+        <Input
```
tasn: You probably want to use TextField and not Input.

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,101 @@
...
+        <RadioGroup
+          row
+          value={ends}
+          onChange={(event: React.FormEvent<{ value: unknown }>) => { setEnds(event.currentTarget.value as string) }}
+        >
+          <FormControlLabel value="never" control={<Radio />} label="Never" />
+          <FormControlLabel value="after" control={<Radio />} label="After" />
+          <FormControlLabel value="onDate" control={<Radio />} label="On Date" />
+        </RadioGroup>
+        <Input
```
tasn: You probably want to use text field, and not Input

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const [count, setCount] = React.useState<number | null | undefined>();
+  const rule = new RRule({
+    freq: frequency,
+    interval: interval,
+    dtstart: new Date(),
+    until: until,
+    count: count,
+  });
+
+  props.sendRRule(rule.toString());
```
tasn: Not the right place to do that.

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,249 @@
...
+  onChange: (rrule: string) => void;
+  rrule: string;
+}
+const frequency = {
+  yearly: 0,
+  monthly: 1,
+  weekly: 2,
+  daily: 3,
+};
+const weekdays = { 0: 'MO', 1: 'TU', 2: 'WE', 3: 'TH', 4: 'FR', 5: 'SA', 6: 'SU' };
```
tasn: This and a few types already exist in rrule, you can just import them (or at least it looks like it):
https://github.com/jakubroztocil/rrule/blob/master/src/weekday.ts#L6

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,249 @@
...
+import { TextField, Select, MenuItem, FormGroup, FormControlLabel, Checkbox } from '@material-ui/core';
+import RRule, { Options, Weekday } from 'rrule';
+
+import DateTimePicker from '../widgets/DateTimePicker';
+
+interface PropsType {
+  onChange: (rrule: string) => void;
+  rrule: string;
+}
+const frequency = {
```
tasn: This type also exists.
https://github.com/jakubroztocil/rrule/blob/master/src/types.ts#L13

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,249 @@
...
+  yearly: 0,
+  monthly: 1,
+  weekly: 2,
+  daily: 3,
+};
+const weekdays = { 0: 'MO', 1: 'TU', 2: 'WE', 3: 'TH', 4: 'FR', 5: 'SA', 6: 'SU' };
+const months = {
```
tasn: Same with this, it probably exists somewhere, but anyhow, you can at least make it an enum which would be much simpler/cleaner:

enum Months {
    Jan = 1,
    Feb,
    Mar,
    ...
}

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,249 @@
...
+  );
+});
+
+
+export default function RRuleEteSync(props: PropsType) {
+  const options = RRule.fromString(props.rrule).origOptions;
+
+  const updateRule = (newOptions: Partial<Options> = {}, weekdayToRemove: Weekday | null = null) => {
+    const updatedOptions = { ...options, ...newOptions };
+    if (Array.isArray(options.byweekday) && Array.isArray(newOptions.byweekday)) {
```
tasn: Weekday to remove is a hack and shouldn't exist. And to an extent also this check with options.byweekday if it's an array.

Instead of weekday to remove, you should just pass the new array from the outside of this function. Intsead of isArray, just pass the new array to be used, like you would any other value.

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,249 @@
...
+    } else if (weekdayToRemove && Array.isArray(updatedOptions.byweekday)) {
+      updatedOptions.byweekday = updatedOptions.byweekday.filter((day) => {
+        return day.toString() !== weekdayToRemove.toString();
+      });
+    }
+    const newRule = new RRule(updatedOptions);
+    props.onChange(newRule.toString());
+  };
+
+  const getEnds = () => {
```
tasn: This is a small stylistic choice, but the rest of the code uses normal functions rather than const = arrow.

So change this to
`function getEnds() {`
if you can

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,249 @@
...
+      {options.bysetpos &&
+        <Select
+          value={options.bysetpos}
+          onChange={(event: React.FormEvent<{ value: unknown }>) => {
+            updateRule({ bysetpos: Number((event.target as HTMLInputElement).value) });
+          }}>
+          <MenuItem value={1}>First</MenuItem>
+          <MenuItem value={2}>Second</MenuItem>
+          <MenuItem value={3}>Third</MenuItem>
+          <MenuItem value={4}>Foruth</MenuItem>
```
tasn: Type, should be Fourth I assume?

tasn: Should it be a set of values like this though? Not an arbitrary number input? I'm asking, I have no idea.

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,249 @@
...
+        return day.toString() !== weekdayToRemove.toString();
+      });
+    }
+    const newRule = new RRule(updatedOptions);
+    props.onChange(newRule.toString());
+  };
+
+  const getEnds = () => {
+    if (options.until && !options.count) {
+      return 'onDate';
```
tasn: Shouldn't the return values in this function and the next be an enum rather than strings?

src/widgets/RRule.tsx
tasn: Leave it a selection like this for now actually, don't worry about arbitrary number inputs.

src/widgets/RRule.tsx
tasn:
```
const frequency = [ Frequency.YEARLY, Frequency.MONTHLY, Frequency.WEEKLY, Frequency.DAILY ];
```

src/widgets/RRule.tsx
Tal-Leibman: ok ill change all of them

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,284 @@
+import * as React from 'react';
+import Container from './Container';
+import { TextField, Select, MenuItem, FormGroup, FormControlLabel, Checkbox, InputLabel, FormControl } from '@material-ui/core';
+import RRule, { Options, Weekday, Frequency, ByWeekday } from 'rrule';
+import { ALL_WEEKDAYS } from 'rrule/dist/esm/src/weekday';
```
tasn: You don't want to import like this. If it's not exposed using a recommended path it probably means it's internal and shouldn't be imported...

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,284 @@
...
+  May,
+  Jun,
+  Jul,
+  Aug,
+  Sep,
+  Oct,
+  Nov,
+  Dec,
+}
+enum MonthReapet {
```
tasn: The naming is inconsistent when all is lower.

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,284 @@
...
+  bysetpos,
+  bymonthday,
+}
+enum Position {
```
tasn: This doesn't need to be an enum, it's really just a numerical value, it could also be 5, that's an allowed value. We just don't expose it in the UI.

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,284 @@
...
+  First = 1,
+  Second = 2,
+  Third = 3,
+  Fourth = 4,
+  Last = -1,
+}
+const menueItemsPosition = Object.keys(Position).filter((key) => Number(key)).map((key) => {
```
tasn: Here and everywhere else: it should be "menu" not "menue" :)

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,284 @@
...
+  return (
+    <MenuItem key={index} value={index}>{value}</MenuItem>
+  );
+});
+const menuItemsFrequency = Object.keys(Frequency).filter((key) => Number(key) < 4).map((value) => {
+  return (
+    <MenuItem key={value} value={value}>{Frequency[value].toLowerCase()}</MenuItem>
+  );
+});
+const menuItemDaysNumbers = Array.from(Array(31).keys()).map((value: number) => {
```
tasn: Having 31 menu items is ugly. Better make it a free field that people can just enter a number. Also, I think the numbers are not mutually exclusive, I don't think there should be just one choice, should there?

src/widgets/RRule.tsx
```diff
@@ -0,0 +1,284 @@
...
+  }
+  function isWeekdayChecked(day: number): boolean {
+    const weekdayArray = options.byweekday as ByWeekday[];
+    if (weekdayArray) {
+      return !!(weekdayArray.find((value) => (value as Weekday).weekday === day));
+    } else {
+      return false;
+    }
+  }
+  const checkboxWeekDays = ALL_WEEKDAYS.map((value, index) => {
```
tasn: Super nit-picky: annoying visually that here it's "WeekDay" and everywhere else is "Weekday". It should be consistent (probably as the latter...)

src/widgets/RRule.tsx
Tal-Leibman: I can change it to TextField with min 1 max 31

Tal-Leibman: bymonthday in rrule can be a list of numbers , what input would make sense for multi select in range  1 to 31?

src/widgets/RRule.tsx
Tal-Leibman: this value "bysetpos" can also be a list of numbers. same question  as "bymonthday"

tasn: There are two relevant questions to be asked:
1. How is the type represented in the code (essentially my question above).
2. How is the selection value represented in the UI.

I'm not asking about the second at the moment, because that can be changed as needed. As for the first, sure, it can be a list of values, but moving from one value to a list (when we support it) is a straightforward change, so it's not a big deal to worry about it now. Though misrepresenting the value sends the wrong message to people reading the code. So if people think it's an enum but it isn't, that's confusing.

src/widgets/RRule.tsx
tasn: Textfield sounds good. As for multiselect in this range: just the ability to add more of these fields (so a + button and an x button, for removal), though I wouldn't worry about it in stage 1.

Have you looked how it's represented in other apps?

Tal-Leibman: image
on google calendar they just don't support all the rrule options just the basic ones

tasn: What about other apps?

tasn:
Looking at the code I don't see any obvious issues. I haven't ran it locally though.
I'm OK with putting it in this weekend. If you can though, please consider doing some basic design. Essentially adding labels and making things a bit more spacious.
https://github.com/Fafruch/react-rrule-generator could be a place to draw inspiration from.

Don't worry about wasting too much screen space vertically, because it's only open when people turn recurrence on, and anyway, being clear is better than being compact (referring to the labels).

Good job!

tasn:
Merged in 6446777c7983e5a8194d34c3499b26283fa0e3e0

PR #70, opened 2020-01-04
widgets DateTimePicker: add disabled property

Tal-Leibman:


tasn:
Nice catch, it's indeed missing. Thanks a lot!

PR #72, opened 2020-01-05
widgets: RRule

Tal-Leibman:


Tal-Leibman:
sorry about that since if talked before i thought it made sense, ill try to explain it better.
Looking at the way we deal with rrule in the IOS 
https://github.com/etesync/ios/blob/master/src/types/ical.js.d.ts#L147;
my enum name values were wrong (Year instead of YEARLY) so i changed that to fit the way we deal with rrule elsewhere since we ended up not using -  https://github.com/jakubroztocil/rrule
as we planned at the start.
the added props are to compatible with 
https://github.com/etesync/ios/blob/master/src/types/ical.js.d.ts#L159

tasn:
Nothing to be sorry about. As I said in my comment I understand what you did there, the code is fine, the commit message is not though, it doesn't explain what you did there at all. Even knowing what the code does and what you intended to do (because we talked in advance), I can't decipher the commit message. :)

Could you please change the commit message and squash together the two fragment commits? Hint: the right tool for the job is `git rebase -i` aka "interactive rebasing". If you don't know how/don't want to figure it out, no worries, let me know and I'll do it.

PR #73, opened 2020-01-11
WIP:  EventEdit: add support for editing event recurrence

Tal-Leibman:


src/components/EventEdit.tsx
```diff
@@ -55,7 +57,7 @@ class EventEdit extends React.PureComponent<PropsType> {
     location: string;
     description: string;
     journalUid: string;
-
```
tasn: This space was put there on purpose. I don't blame you for now realising it, because it's terribly hacky what's going on there, but the idea is that the above are actual properties of the event, and below are just local state.

So please keep the space and put the rrule options (why not just call it rrule?) below timezone.

src/components/EventEdit.tsx
```diff
@@ -69,7 +71,6 @@ class EventEdit extends React.PureComponent<PropsType> {
       location: '',
       description: '',
       timezone: null,
-
```
tasn: Same comment about the space. Again, it's terrible, and will be changed when this component is refactored, though this space keeps some sanity for now.

src/components/EventEdit.tsx
```diff
@@ -154,7 +158,16 @@ class EventEdit extends React.PureComponent<PropsType> {
...
   }
+  public toggleRecurring() {
+    const value = this.state.rruleOptions ? undefined : { freq: 'WEEKLY', interval: 1 };
+    this.setState({ rruleOptions: value });
+  }
+
 
+  public handleRRuleChange(rrule: RRuleOptions): void {
+    this.setState({ rruleOptions: rrule });
+    console.log(rrule);
```
tasn: `console.log`... Please remove. :)
I'd make it a habit not to commit those, not even in a WIP branch.

src/components/EventEdit.tsx
```diff
@@ -190,7 +203,7 @@ class EventEdit extends React.PureComponent<PropsType> {
       this.props.item.clone()
       :
       new EventType()
-    ;
+      ;
```
tasn: Unrelated whitespace changes.

src/components/EventEdit.tsx
```diff
@@ -232,21 +245,17 @@ class EventEdit extends React.PureComponent<PropsType> {
       },
     };
 
-    const recurring = this.props.item && this.props.item.isRecurring();
+    if (this.props.item && this.props.item.isRecurring()) {
+      console.log(1);
```
tasn: Same comment about logs being pushed to PRs.

src/components/EventEdit.tsx
```diff
@@ -232,21 +245,17 @@ class EventEdit extends React.PureComponent<PropsType> {
...
         <h2>
           {this.props.item ? 'Edit Event' : 'New Event'}
         </h2>
-        {recurring && (
-          <div>
-            <span style={{ color: 'red' }}>IMPORTANT: </span>
-            This is a recurring event, for now, only editing the whole series
-            (by editing the first instance) is supported.
-          </div>
-        )}
```
tasn: This warning is still needed and is still valid. We are still only editing the whole event series rather than let people edit specific instances to create recurrence exceptions.

src/components/EventEdit.tsx
```diff
@@ -373,16 +400,17 @@ class EventEdit extends React.PureComponent<PropsType> {
           </div>
         </form>
 
+
```
tasn: Whitespace changes.

src/components/EventEdit.tsx
```diff
@@ -373,16 +400,17 @@ class EventEdit extends React.PureComponent<PropsType> {
...
         <ConfirmationDialog
           title="Delete Confirmation"
           labelOk="Delete"
           open={this.state.showDeleteDialog}
           onOk={() => this.props.onDelete(this.props.item!, this.props.initialCollection!)}
           onCancel={() => this.setState({ showDeleteDialog: false })}
         >
-        Are you sure you would like to delete this event?
+          Are you sure you would like to delete this event?
```
tasn: I'll let this one slide, because this one is actually correct. :)

src/components/EventEdit.tsx
```diff
@@ -33,6 +33,8 @@ import * as EteSync from 'etesync';
 import { getCurrentTimezone } from '../helpers';
 
 import { EventType, timezoneLoadFromName } from '../pim-types';
+import RRuleEteSync, { RRuleOptions } from '../widgets/RRule';
```
tasn: Remove the extra line you added, and probably just call it RRule, whit call it RRuleEteSync?

tasn:
Looks great, awesome job! It probably needs a bit of design touchups and I have a few comments, but it looks great overall.

1. I'd probably hide the more complex rules (day of the month, months, though keep weekdays) until we think on how to show them in the UI. It looks like Android doesn't support them.
2. I'd drop the colon before the interval.
3. I'd change the "Repeat every 1 weekly" text to "Repeat every 1 week", it looks odd.

Again, great job!

src/components/EventEdit.tsx
Tal-Leibman: its RRuleEteSync because the when we started we used the rrule.js library so it had a name conflict with RRule. but now i can change it to just RRule

Tal-Leibman:
>     1. I'd probably hide the more complex rules (day of the month, months, though keep weekdays) until we think on how to show them in the UI. It looks like Android doesn't support them.
months multi select day  is only visible when freq is set to yearly, should i still hide it always?
and by month day is visible only when freq is set to monthly , also its optional.
>     2. I'd drop the colon before the interval.
no problem 
>     3. I'd change the "Repeat every 1 weekly" text to "Repeat every 1 week", it looks odd.
now its just displaying the enum string value ill map it a const {"WEEKLY":"week" ...etc}
> Again, great job!
Thank for all the feedback I am definitely learning a lot :)

src/components/EventEdit.tsx
Tal-Leibman: should i make a single commit with all the linter white space changes? 
if not ill just remove the changes

src/components/EventEdit.tsx
Tal-Leibman: ill move it back to the top like before

src/components/EventEdit.tsx
Tal-Leibman: yeah it definitely didn't get committed on purpose

src/components/EventEdit.tsx
Tal-Leibman: ill change it to just rrule

tasn:
> months multi select day is only visible when freq is set to yearly, should i still hide it always?
> and by month day is visible only when freq is set to monthly , also its optional.

I'm on the fence here. The problem is that I think seeing them will confuse people. Let's just leave it as is for now. Though maybe when none are selected (which means all, no?) just have the preview say something like "all"? I'm not sure. Happy to punt this decision to later.

src/components/EventEdit.tsx
tasn: I don't think this one was added by the linter, did it? Though anyway, even if it was, don't add unrelated whitespace changes to commits.

src/components/EventEdit.tsx
tasn: Gotcha. Yeah, please fix it.

src/components/EventEdit.tsx
tasn: rrule need to be before the space, not after. More specifically, I'd probably put it even before journalUid which I think is also not really part of the core event state.

tasn:
To update the rrule property:
```
    const value = rruleNativeToVobject(event.recurrenceRule, event.allDay);
    ret.component.addPropertyWithValue('rrule', value);
```

As seen can be seen in https://github.com/etesync/ios/blob/master/src/sync/helpers.ts#L275

Issue #9, opened 2018-02-16
Feature request: change date and time picker


davidwisniewski:
The current date and time picker is pretty clunky. Picking the day is fine, I can just select it from a calendar. But in order to pick a start and end time, I need to scroll through hours and minutes every time. 
It would help if e.g. there would be a default duration at least (1h?), so that if I pick the start time, the end time is added automatically. Or, give the option to type in date and time manually with the keyboard, which would be faster. Even better, switch to a different time picker that is more intuitive to use (e.g. clock icon).

tasn:
Thanks for reporting this.
The current date and time picker is horrible. It's pretty bad on mobile, and horrific on the desktop. I'd love to change it back to the native pickers in browsers, but unfortunately they don't work well on all
browsers, which is why I changed it to this in the first place. I know it needs to be changed.

Would need to take a second look into using the native browser picker and having this as a fallback.

rugk:
Hmm, [about 80%](https://caniuse.com/#feat=input-datetime), but Firefox does not support combinations (only date or time) – maybe just use two input fields?

tasn:
@rugk, it was actually 2 input fields because of Firefox's poor support. Though I don't remember who reported it, but the problem was another browser. Maybe Chrome mobile, I'm not sure...

tasn:
@Tal-Leibman, maybe this is also of interest? I can't believe it took me so long to revisit this. :| Sorry everyone.

PR #74, opened 2020-01-17
Adds some initial improvements to Tasks

AbleLincoln:
This is the MVP for a few features I want to add to Tasks. The code is by no means good or clean yet, I just wanted to get out a few rough ideas first. Let me know what you think of the following. Anything can be changed.

I've added:
- quick input right from the list screen
- tags: They show up next to tasks and can be added via the edit form. The textbox currently accepts space-separated values for the tags. I would like to improve this input method.
- edit form cleanup: there were a lot of fields that I felt were extraneous (location, timezone, start date) so I have removed them. Let me know if you actually want to keep any of these
- better datetimepicker. Material UI has a pretty good component for this that we were not using, so I have dropped theirs in.


src/pim-types.ts
```diff
@@ -171,6 +171,14 @@ export class TaskType extends EventType {
     return !!((this.startDate?.isDate) || (this.dueDate?.isDate));
   }
 
+  set tags(tagHash: any) {
+    this.component.updatePropertyWithValue('tags', JSON.stringify(tagHash));
+  }
+
+  get tags() {
+    return JSON.parse(this.component.getFirstPropertyValue('tags'));
```
tasn: I'll properly review everything slightly later, when you make more progress, but one important comment: this shouldn't be called "tags" but "properties", and you shouldn't JSON.stringfiy it, but rather it should be an array of strings.

This is not something that comes from me, it comes from the [ical standard](https://tools.ietf.org/html/rfc5545#section-3.8.1.2). This object that you're updating is an ical object that then needs to be serialized to the ical format. It doesn't change much, just make sure to rename it and change the value. Additional properties of tags, such as colour, should be saved elsewhere, and I'll happily explain it when you get to saving those. Though those should be maintained at a per-journal capacity as part of the journal metadata.

src/components/TaskList.tsx
```diff
@@ -42,21 +87,18 @@ export default React.memo(function TaskList(props: PropsType) {
...
     const uid = entry.uid;
 
-    return (
-      <TaskListItem
-        key={uid}
-        entry={entry}
-        onClick={props.onItemClick}
-      />
-    );
+    return <TaskListItem key={uid} entry={entry} onClick={props.onItemClick} onSave={props.onItemSave} />;
```
tasn: Actually, another small comment that is better said now. This is one example (but I saw it a few times in this patch). Please don't change the formatting when you don't need to and just conform to what's already there. For example, here you changed the formatting which makes the diff unnecessarily long and makes it harder than it needs to be to see that you just added a property (on save) and passed it.

package.json
```diff
@@ -3,10 +3,14 @@
...
   "private": true,
   "dependencies": {
+    "@date-io/date-fns": "1.x",
+    "@date-io/moment": "^2.0.1",
     "@material-ui/core": "^4.7.0",
     "@material-ui/icons": "^4.5.1",
     "@material-ui/lab": "^4.0.0-alpha.33",
+    "@material-ui/pickers": "^3.2.8",
     "@material-ui/styles": "^4.6.0",
+    "date-fns": "^2.9.0",
```
tasn: Actually, one more note about deps:
date-fns: please don't use it. We already have moment and use that, and it's compatible with material-ui/pickers, so really no need to have it. Same with @date-io/date-fns.

In general, please favour using existing deps to adding new ones.

src/pim-types.ts
AbleLincoln: I see. I think I've got it now. It was a little confusing to figure out how to do multi value properties, but looks like I need to explicitly add the "CATEGORIES" property to the `TaskType`.

AbleLincoln: > Additional properties of tags, such as colour, should be saved elsewhere, and I'll happily explain it when you get to saving those. Though those should be maintained at a per-journal capacity as part of the journal metadata.

Could I also save a list of all the tags this way? If so, how would I go about that?

src/components/TaskEdit.tsx
```diff
@@ -92,6 +89,8 @@ class TaskEdit extends React.PureComponent<PropsType> {
       this.state.location = event.location ? event.location : '';
       this.state.description = event.description ? event.description : '';
       this.state.timezone = event.timezone;
+      console.log(event);
+      this.state.tags = event.categories ? event.categories.join(' ') : '';
```
tasn: I know it's probably way too early too comment on this, so feel free to ignore, but I'd rather not have free text it, but rather a sort of a way to select existing tags (multiselect popup?). Maybe in addition we can also have free text (comma delimited though, not space), but I feel like it would need a nice autocomplete, so probably easier to just defer that to a later stage.

src/pim-types.ts
tasn: The list of tags is external to specific tasks. It should probably be stored in each journal's metadata. I'm happy to explain to you how to do it (it would require a tiny bit of code on my side), though I think it's too early for that, just save it locally (and globally) for now, or heck, make the tags hardcoded while you are working on it, and ping me again in a few days / a week and I'll add the ability to set generic journal metada.

src/components/TaskList.tsx
tasn: ESLint is not complaining here, so I don't think it's eslint, I think you have some additional rules in your VSCode or whatever IDE you are using...

tasn:
Hey,

Just one general comment, if you fix a general issue with the code, please have it in a separate PR and at the very least a separate commit in a PR so it can be easily merged. I just fixed #9, and had to implement it from scratch because of this. A fix that could have been in for a while now... :)

Additionally, your fix was incorrect because it was ignoring the "dateOnly" prop.

AbleLincoln:
I've added a bunch of features:

## Task properties
- tags: currently tags are hard-coded, we have "work" and "home". I will add dynamic tags in a follow-up PR, it shouldn't be too hard
- priority: none, low, med, high

## Task list
- colored check boxes corresponding to priority
- tags
- secondary text displaying how long until the task is due

## Sidebar
Basically where all of the filtering happens. Currently you can filter by all, due today, and tag

## Toolbar
Contains UI for sorting, searching, and toggling show completed. Currently you can sort by priority, due date, title, last modified.

## Context
I am using a bit of React Context to pass down props for saving and selecting tasks

## Redux Store
The store now holds settings for the task page. This allows the user to maintain the filters, sorts, etc. between page reloads.

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "outDir": "build/dist",
     "module": "esnext",
-    "target": "es5",
+    "target": "es6",
```
tasn: Why did you change that? This change is not desired, we were targeting es5 on purpose.

src/types/ical.js.d.ts
```diff
@@ -43,7 +43,7 @@ declare module 'ical.js' {
 
     public component: Component;
 
-    public constructor(component?: Component | null, options?: {strictExceptions: boolean, exepctions: Array<Component | Event>});
+    public constructor(component?: Component | null, options?: { strictExceptions: boolean, exepctions: Array<Component | Event> });
```
tasn: All of the changes to this file should be removed, please don't mix style and logic changes in a PR, it makes things very hard to review.

src/store/reducers.ts
```diff
@@ -291,15 +291,32 @@ export const errorsReducer = handleActions(
...
+      { ...state, ...action.payload }
     ),
   },
-  { locale: 'en-gb' }
+  {
+    locale: 'en-gb',
+
+    tasks: {
+      showCompleted: false,
+      sortOrder: 'dueDate',
```
tasn: I'm not sure if you took a look at the code I sent you back then, but I assume you have some sort of a "smart" sort, that should be the default, not this.

src/store/reducers.ts
```diff
@@ -291,15 +291,32 @@ export const errorsReducer = handleActions(
 
 
 // FIXME Move all the below (potentially the fetchCount ones too) to their own file
+export interface TaskSettings {
+  showCompleted: boolean;
+  sortOrder: string;
+  filterBy: string;
+  searchTerm: string;
```
tasn: This should not be a setting, it's ephemeral, not persistent... It should just be a prop / navigation property.

src/Pim/index.tsx
```diff
@@ -44,14 +44,14 @@ function objValues(obj: any) {
 }
 
 const itemsSelector = createSelector(
-  (props: {syncInfo: SyncInfo}) => props.syncInfo,
+  (props: { syncInfo: SyncInfo }) => props.syncInfo,
```
tasn: Please don't mix style and logic changes in a PR, it makes things very hard to review. Same goes for almost all of the rest of this file.

src/Pim/index.tsx
```diff
@@ -115,10 +115,10 @@ type CollectionRoutesPropsType = RouteComponentProps<{}> & {
 
 const styles = (theme: any) => ({
   button: {
-    marginLeft: theme.spacing.unit,
+    marginLeft: theme.spacing(1),
```
tasn: This is a fine change, as unit is deprecated for all I remember, but this change is unrelated to your PR. Please don't mix unrelated changes. I'd be happy to merge this if you open a different PR with this fix.

src/Pim/index.tsx
```diff
@@ -220,6 +220,11 @@ const CollectionRoutes = withStyles(styles)(withRouter(
   }
 ));
 
+export const PimContext = React.createContext({
+  onItemSave: (_item: PimType, _journalUid: string, _originalEvent?: PimType, _goBack = true) => { return },
+  collectionsTaskList: [] as EteSync.CollectionInfo[],
+});
+
```
tasn: Context does not seem like the correct tool for this job... Why can't it be passed as props? I must be missing something, but I don't understand what you are trying to do.
Also, have tasks specific code in an otherwise generic file is a code smell.

src/Pim/index.tsx
```diff
@@ -235,7 +240,7 @@ class Pim extends React.PureComponent {
     this.onItemSave = this.onItemSave.bind(this);
   }
 
-  public onItemSave(item: PimType, journalUid: string, originalEvent?: PimType) {
+  public onItemSave(item: PimType, journalUid: string, originalEvent?: PimType, goBack = true) {
```
tasn: I don't see anywhere where you used this extra parameter, nor do I understand why it's needed. Could you please elaborate?
If I had to assume, I guess you did it for marking items as "done" in the task list page without going back, though I can't find where it's used.

Additionally, rather than have this extra parameter, I think it would be cleaner to remove the history.goback altogether from this code, make it return a promise, and then whichever piece of code that needs to go back after the save is done, can just have it there (probably just in contactedit, eventedit, and taskedit). What do you think?

src/components/Tasks/AddNewTaskItem.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setTitle(e.target.value);
+  };
+
+  const handleKeyPress = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === 'Enter') {
+      e.preventDefault();
+
+      const event = new TaskType(null);
```
tasn: A bit of a nit-pick, but looks like a bad copy-paste. The variable should be called "task" as it's a task. :)

src/components/Tasks/AddNewTaskItem.tsx
```diff
@@ -0,0 +1,50 @@
...
+      event.uid = uuid.v4();
+      event.title = title;
+      event.status = TaskStatusType.NeedsAction;
+      event.priority = TaskPriorityType.None;
```
tasn: Is setting status+priority needed? Aren't these the default? I really have no idea and just asking.

src/components/Tasks/AddNewTaskItem.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { useState, useContext } from 'react';
```
tasn: See the rest of the code, we follow a different convention.
We do `import * as React from 'react'` and then use `React.useState` in the code.

src/components/Tasks/AddNewTaskItem.tsx
```diff
@@ -0,0 +1,50 @@
...
+      e.preventDefault();
+
+      const event = new TaskType(null);
+      event.uid = uuid.v4();
+      event.title = title;
+      event.status = TaskStatusType.NeedsAction;
+      event.priority = TaskPriorityType.None;
+      event.lastModified = ICAL.Time.now();
+
+      save(event, collections[0].uid, undefined, false);
```
tasn: It's missing error-handling. What if the save fails?

src/components/Tasks/ColoredRadio.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
```
tasn: Same comment as before, we use a slightly different convention: `import * as React from 'react';`

src/components/Tasks/ColoredRadio.tsx
```diff
@@ -0,0 +1,32 @@
...
+import FormControlLabel, { FormControlLabelProps } from '@material-ui/core/FormControlLabel';
+
+interface Props {
+  color: string;
+  label: string;
+}
+
+const useStyles = makeStyles({
+  root: {
+    color: (props: Props) => colors[props.color][600],
```
tasn: Why do this rather than just have color be the actual color? As in `#RRGGBBAA`? You can then pass the correct color values from the outside. After you do this, it also becomes a more generic widget and should therefore be moved under `widgets`.

src/components/Tasks/Sidebar.tsx
```diff
@@ -0,0 +1,81 @@
...
+      key={tag}
+      name={tag}
+      primaryText={tag}
+      secondaryText={String(amount)}
+      Icon={LabelIcon} />
+  ));
+
+  return (
+    <List dense>
+      {filterBy === 'search' && <SidebarListItem name="search" primaryText={`Search "${searchTerm}"`} />}
```
tasn: This feels wrong to me. "Search" shouldn't be a filtering option, because you should be able to e.g. "show/hide deleted" + search at the same time. This can easily be solved by having search be a special page (I made another comment somewhere that shared my thoughts on this).

src/components/Tasks/TaskEdit.tsx
```diff
@@ -11,28 +13,20 @@ import MenuItem from '@material-ui/core/MenuItem';
...
 import IconCancel from '@material-ui/icons/Clear';
 import IconSave from '@material-ui/icons/Save';
 
-import DateTimePicker from '../widgets/DateTimePicker';
-
-import ConfirmationDialog from '../widgets/ConfirmationDialog';
-import TimezonePicker from '../widgets/TimezonePicker';
-
-import { Location } from 'history';
-import { withRouter } from 'react-router';
```
tasn: All of this unnecessary reordering makes it impossible to review. Please revert this and only change what you are actually changing.

src/components/Tasks/TaskEdit.tsx
```diff
@@ -92,6 +89,10 @@ class TaskEdit extends React.PureComponent<PropsType> {
       this.state.location = event.location ? event.location : '';
       this.state.description = event.description ? event.description : '';
       this.state.timezone = event.timezone;
+      this.state.priority = event.priority ? event.priority : TaskPriorityType.None;
```
tasn: I know the rest of the code uses the old syntax (because it's a fairly new typescript feature), but you can now just do: `this.state.priority = event.priority ?? TsakPrioritiyType.None`

src/components/Tasks/TaskEdit.tsx
```diff
@@ -92,6 +89,10 @@ class TaskEdit extends React.PureComponent<PropsType> {
...
+
+      // checks if event categories is just the empty string
+      this.state.tags = event.categories && event.categories[0] ? event.categories : [];
```
tasn: Please enclose the condition in parenthesis, it's very confusing otherwise.
Additionally, if you are just checking if event.categories is an empty string, why not do `event.categories === ''`?

src/components/Tasks/TaskEdit.tsx
```diff
@@ -359,8 +335,8 @@ class TaskEdit extends React.PureComponent<PropsType> {
             {this.props.item &&
               <Button
                 variant="contained"
-                style={{ marginLeft: 15, backgroundColor: colors.red[500], color: 'white' }}
-                onClick={this.onDeleteRequest}
+                style={{ marginLeft: 15, backgroundColor: red[500], color: 'white' }}
+                onClick={() => this.onDeleteRequest()}
```
tasn: Why this change?

src/components/Tasks/TaskEdit.tsx
```diff
@@ -391,7 +362,7 @@ class TaskEdit extends React.PureComponent<PropsType> {
           onOk={() => this.props.onDelete(this.props.item!, this.props.initialCollection!)}
           onCancel={() => this.setState({ showDeleteDialog: false })}
         >
-        Are you sure you would like to delete this event?
+          Are you sure you would like to delete this event?
```
tasn: Please, no unrelated styling changes. Additionally, please change "event" to "task".

src/components/Tasks/TaskEdit.tsx
```diff
@@ -277,7 +272,7 @@ class TaskEdit extends React.PureComponent<PropsType> {
             <Select
               name="status"
               value={this.state.status}
-              onChange={this.handleInputChange}
+              onChange={(e) => this.handleInputChange(e)}
```
tasn: Why make this change? (+ the 3 ones before it)?

src/components/Tasks/TaskListItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+      key={i}
+      color="secondary"
+      size="small"
+      label={tag}
+      style={{ marginRight: '10px' }}
+      component="li"
+    />)}
+  </ul>));
+
+// FIXME: HACK the default colors just happen to work, probably want hand-picked colors
```
tasn: What do you mean by this?

src/pim-types.ts
```diff
@@ -80,6 +80,10 @@ export class EventType extends ICAL.Event implements PimType {
     return this.summary;
   }
 
+  set title(_title: string) {
```
tasn: The underscore prefix convention is for unused parameters, so shouldn't be used here.

Note: I know it's used elsewhere in the code for something else, but still, nothing to be copied, and was never just used like this.

src/pim-types.ts
```diff
@@ -184,6 +199,30 @@ export class TaskType extends EventType {
     return !!((this.startDate?.isDate) || (this.dueDate?.isDate));
   }
 
+  set categories(tags: string[]) {
+    this.component.getFirstProperty('categories').setValues(tags);
```
tasn: Are you sure this works as expected, as in, does the resulting file look like it follows the spec: https://tools.ietf.org/html/rfc5545#section-3.8.1.2 ?
I'm asking because I've had issues with setValues before.

src/store/actions.ts
```diff
@@ -223,7 +223,7 @@ export const clearErros = createAction(
 // FIXME: Move the rest to their own file
 export const setSettings = createAction(
   'SET_SETTINGS',
-  (settings: SettingsType) => {
+  (settings: Partial<SettingsType>) => {
```
tasn: I'm OK with this change, though again, like many of my other comments, this is a very easy to miss change that is quite wide-reaching. Please make these in different PRs in the future, or at the very least different commits!

tsconfig.json
AbleLincoln: That's my bad. I think I changed this to use some ES6 only code, but I have since removed the code. I will change this back

src/components/Tasks/TaskEdit.tsx
AbleLincoln: Sorry, my comment is poorly written. It should say `checks if event categories contains only the empty string`.
iCAL is kind of silly in how it handles empty lists. If, for instance, you have a task with no tags, its _categories_ property will be saved with the value `['']`.

For backwards compatibility, I first check to see if the task even has a _categories_ property.

This code is a little misleading though. Perhaps it can be written in a cleaner way.

tasn: See my other comment about the usage of `getValues`, I think it's related. I think it actually shouldn't be getValues, but rather a comma (or is it semicolon?) delimited string that you should deal with on your own. I really don't remember, but I know I've had issues with it before.

tasn:
One more related comment to save both of us a lot of time: maybe let's do things step by step? So maybe have a PR that I can merge that changes the things that are not really related to this PR? And then have another small PR that does X and I can already merge, and so on? So we at least know things are good and can move forward? Only if it makes things easier for you...

What I'm essentially trying to do is shorten the feedback loop, so you can change things and get immediate feedback rather than sitting on your changes for a few days, waiting for my feedback, and then sitting on your changes again for a few days.

AbleLincoln:
Yes I think that is a good idea. What would be the best way to go about this regarding git/GitHub since I already have a lot of these changes in progress on this branch. I'm sorry, I'm just not too experienced with complicated merges.

tasn:
> Yes I think that is a good idea. What would be the best way to go about this regarding git/GitHub since I already have a lot of these changes in progress on this branch.

Great, so I'm looking forward to your smaller PRs! Just think what changes make sense standalone and open the appropriate PRs. The added benefit is that we can discuss separate topics separately. So for example, we will be able to discuss search in one and quick-add in another.

> I'm sorry, I'm just not too experienced with complicated merges.

Don't be silly, nothing to be sorry about! I didn't mean my previous messages as complaints, I meant them as guiding. I know it's new for you, and I'm here to help! I was just telling you what I think could make things easier based on my experience.

src/store/reducers.ts
AbleLincoln: The reason I have these as settings is that I want the filtering/sorting to persist when the user navigates away from the Tasks page. The main use case is when a user goes to edit/create a task. After saving and navigating back to the tasks page, all the sorting/filtering is reset, which can be frustrating.

Do you have a suggestion for another way I can handle this?

AbleLincoln:
Thank you for all your help and understanding. Do you recommend I just create completely new branches from the master and copy over changes?

src/Pim/index.tsx
AbleLincoln: I am pretty much just using this for the task "quick add" feature. Without that parameter, the app takes you back to the homepage every time you try to add a task. But what I want is to be able to stay on the same page after saving. 

I know this solution seems silly, but I did it to change the least amount of code possible. I agree that a promise would make much more sense. Would require editing some other code. I can handle that in a separate PR

src/store/reducers.ts
tasn: Sorry if my initial comment wasn't clear, I was referring specifically to searchTerm, not the rest. The rest should indeed be as you did them. As I told you elsewhere, it's similar to how I did it in etesync-tasks, which you can refer to if you want a glimpse into my thoughts process on things, and anyhow, a lot of it is very similar. Obviously my thoughts aren't necessarily good, but at least you can see what I mean. :)

src/Pim/index.tsx
tasn: As I wrote, I understand why you did it, I just couldn't find anywhere you called this function with this extra parameter. Maybe I missed something because of all the code I had to review, but I don't remember seeing anything where it's used.

Anyhow, I think the promise is a better way of doing it, and how it should have been done in the first place. As I told you, the current code is less than awesome in some parts. This is one of them.

tasn:
Yeah, maybe just creating new branches and copying over changes is the easiest way to go. It's likely that some of your PRs will rely on others, which is fine too. Even within PRs, split to commits as it makes sense to you (maybe no split is needed), so unrelated code isn't grouped together.

tasn:
Hey,

Just one thing I noticed now that we have merged #81: you can't access the change history anymore. That link should be added to the task edit page. Please keep that in mind.

AbleLincoln:
> Just one thing I noticed now that we have merged #81: you can't access the change history anymore. That link should be added to the task edit page. Please keep that in mind.

Noted, thank you for the heads up. I will add this in

tasn:
Another comment that comes to mind: the mapping of High, Medium and Low priority from the UI to ical should be:
High: 1-4
Medium: 5
Low: 6-9

I took a look at the RFC and that's the mapping they used there. Just FYI.

src/components/Tasks/AddNewTaskItem.tsx
AbleLincoln: I believe priority is automatically set (to 0), but status is not. Should I set the status in the constructor?

src/components/Tasks/AddNewTaskItem.tsx
AbleLincoln: This is true; however, there is not really error handling in any part of the app (at least in terms of saving items). Can I save this for a future PR and implement it app wide?

src/components/Tasks/AddNewTaskItem.tsx
tasn: Sorry, let me clarify what I meant: you don't have to set all the fields. Specifically, don't set fields that haven't been changed from the default. The reason for that is that if you don't set fields, they don't show in the resulting ical, which is a good thing. No need to add noise there.

src/components/Tasks/AddNewTaskItem.tsx
tasn: I think item save has error handling, no? Anyhow, I see you took it out to its own PR, I'll comment there once I review it if I have anything more to add. Though if there's no error-handling anywhere else, I'm OK with not adding it here for now.

tasn:
Another thing I noticed, especially given that we now have quickadd: we need a way to choose the default task list - so it'll both be the one in the new event dialog, but also the one added to from quickadd.

src/pim-types.ts
AbleLincoln: I've tested it in the sandbox and it seems to work fine. Any values set with `setValues` are basically converted to a comma separated string when outputted in the resulting ical string

tasn: Let's talk about it in the other PR, though from the looks of it there you didn't use setValues but used the manual comma delimiting. Contrary to what you said here.

AbleLincoln:
closing because all of these features have now been implemented in other PRs

tasn:
\o/

PR #81, opened 2020-02-28
Clicking a task goes directly to edit page

AbleLincoln:
Changes the user flow so that selecting a task bypasses the "Task view" and goes directly to the edit page.

### DIFF
--- a/src/Pim/PimMain.tsx
+++ b/src/Pim/PimMain.tsx
@@ -64,7 +64,7 @@ class PimMain extends React.PureComponent<PropsType> {
     const itemUid = `${(event as any).journalUid}|${event.uid}`;
 
     this.props.history!.push(
-      routeResolver.getRoute('pim.tasks._id', { itemUid }));
+      routeResolver.getRoute('pim.tasks._id.edit', { itemUid }));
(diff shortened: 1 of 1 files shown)


tasn:
Looks good to me, just one comment, could you make the commit messages conform to the rest of the commit messages? I know there's no style guideline at the moment (I should make one!) but you can see from existing messages. E.g. for this one, I'd make it: "Tasks: change clicking on task to go directly to edit page."

AbleLincoln:
Yeah whoops, I kind of messed that up haha. Give me a sec, I can fix

tasn:
Cool. :)
Another cool git tool in case you are unfamiliar with it is: `git rebase -i` which lets you edit history. So you can for example, in this case run `git rebase -i HEAD~4` which will let you squash together the last few commits, reorder them, reword them and whatever.

AbleLincoln:
Does this work?

tasn:
Yup, good job! Sorry again for all of the hassle, but there's quite a bit of code involved and quite a few commits, so I'd rather we do things right from the start, even with these small changes. :)

Issue #32, opened 2019-01-26
Tasks support


ajvsol:
There's not many CalDAV web clients I've seen with tasks support to take inspiration from except for Nextcloud Tasks:
![](https://raw.githubusercontent.com/nextcloud/tasks/master/screenshots/tasks-1.png)

But there's good UIs also from other to-do list apps. I stitched together images of Astrid Tasks' previous layout with the current fork's Android UI (it's called Tasks) to make an appropriate UI for a web app:
![](https://i.imgur.com/wglGRIr.png)

The Tasks app also is compatible with CalDAV so it could make a good foundation for a future EteSync Android app. Would make sense to do a similar UI if so.

As far as featureset goes the CalDAV spec covers basically everything you'd need a MVP tasks app.

tasn:
I pushed some code to handle tasks today. It lets you view and edit tasks.
Though the UI is very bad still... It shouldn't take too much work to get it to the level of the Android example from above, I just need to get to it.

ajvsol:
Been using Orgzly recently and came across [this other issue](https://github.com/etesync/android/issues/62) where it sounds like we'll be getting Orgzly support for Android EteSync soon, which is great to hear! 

It did remind me however that the EteSync Web tasks app is lacking some love right now, and I'm not sure whether it's a bugged default journal but I can't seem to add tasks at all right now via the web app.

tasn:
You can't add from the journal page, but if you go to the main page and go to the tasks tab, you can add just fine. It needs fixing. I know... :)

tasn:
Btw, I don't know if you noticed, but there's been a lot of work recently by @AbleLincoln on this and new features are coming in by the day. We will hopefully be able to close this ticket soon! :)

ajvsol:
Good to see how far it's come, it's a huge difference. I've given some feedback in separate issues here, I hope it's helpful.

tasn:
Yeah, kudos to @AbleLincoln!

PR #84, opened 2020-03-07
Tasks: adds Quick Add feature

AbleLincoln:
Adds a feature to quickly add a new task straight from the tasks list

### DIFF
--- a/src/Pim/PimMain.tsx
+++ b/src/Pim/PimMain.tsx
@@ -10,6 +10,8 @@ import { Theme, withTheme } from '@material-ui/core/styles';
 
 import * as ICAL from 'ical.js';
 
+import * as EteSync from 'etesync';
+
 import { Location, History } from 'history';
 
 import Container from '../widgets/Container';
@@ -18,7 +20,7 @@ import SearchableAddressBook from '../components/SearchableAddressBook';
 import Calendar from '../components/Calendar';
 import TaskList from '../components/Tasks/TaskList';
 
-import { EventType, ContactType, TaskType } from '../pim-types';
+import { EventType, ContactType, TaskType, PimType } from '../pim-types';
 
 import { routeResolver } from '../App';
 
@@ -37,6 +39,8 @@ interface PropsType {
   location?: Location;
   history?: History;
   theme: Theme;
+  collectionsTaskList: EteSync.CollectionInfo[];
+  onItemSave: (item: PimType, journalUid: string, originalContact?: PimType) => void;
 }
 
 class PimMain extends React.PureComponent<PropsType> {
@@ -146,7 +150,9 @@ class PimMain extends React.PureComponent<PropsType> {
           {tab === 2 &&
             <TaskList
               entries={this.props.tasks}
+              collections={this.props.collectionsTaskList}
               onItemClick={this.taskClicked}
+              onItemSave={this.props.onItemSave}
             />
           }
         </Container>
--- a/src/Pim/PimMain.tsx
+++ b/src/Pim/PimMain.tsx
@@ -40,7 +40,7 @@ interface PropsType {
   history?: History;
   theme: Theme;
   collectionsTaskList: EteSync.CollectionInfo[];
-  onItemSave: (item: PimType, journalUid: string, originalContact?: PimType) => void;
+  onItemSave: (item: PimType, journalUid: string, originalItem?: PimType) => void;
 }
 
 class PimMain extends React.PureComponent<PropsType> {
--- a/src/Pim/index.tsx
+++ b/src/Pim/index.tsx
@@ -110,7 +110,7 @@ type CollectionRoutesPropsType = RouteComponentProps<{}> & {
   componentEdit: any;
   componentView: any;
   items: {[key: string]: PimType};
-  onItemSave: (item: PimType, journalUid: string, originalItem?: PimType) => Promise<any>;
+  onItemSave: (item: PimType, journalUid: string, originalItem?: PimType) => Promise<void>;
   onItemDelete: (item: PimType, journalUid: string) => void;
   onItemCancel: () => void;
   classes: any;
@@ -240,7 +240,7 @@ class Pim extends React.PureComponent {
     this.onItemSave = this.onItemSave.bind(this);
   }
 
-  public onItemSave(item: PimType, journalUid: string, originalEvent?: PimType): Promise<History<any>> {
+  public onItemSave(item: PimType, journalUid: string, originalEvent?: PimType): Promise<void> {
     const syncJournal = this.props.syncInfo.get(journalUid);
 
     if (syncJournal === undefined) {
--- a/src/Pim/index.tsx
+++ b/src/Pim/index.tsx
@@ -110,7 +110,7 @@ type CollectionRoutesPropsType = RouteComponentProps<{}> & {
   componentEdit: any;
   componentView: any;
   items: {[key: string]: PimType};
-  onItemSave: (item: PimType, journalUid: string, originalItem?: PimType) => void;
+  onItemSave: (item: PimType, journalUid: string, originalItem?: PimType) => Promise<History<any>>;
   onItemDelete: (item: PimType, journalUid: string) => void;
   onItemCancel: () => void;
   classes: any;
@@ -238,11 +238,11 @@ class Pim extends React.PureComponent {
     this.onItemSave = this.onItemSave.bind(this);
   }
 
-  public onItemSave(item: PimType, journalUid: string, originalEvent?: PimType) {
+  public onItemSave(item: PimType, journalUid: string, originalEvent?: PimType): Promise<History<any>> {
     const syncJournal = this.props.syncInfo.get(journalUid);
 
     if (syncJournal === undefined) {
-      return;
+      return Promise.reject();
     }
 
     const journal = syncJournal.journal;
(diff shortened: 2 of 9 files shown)


src/Pim/PimMain.tsx
```diff
@@ -34,6 +36,8 @@ interface PropsType {
   location?: Location;
   history?: History;
   theme: Theme;
+  collectionsTaskList: EteSync.CollectionInfo[];
+  onItemSave: (item: PimType, journalUid: string, originalContact?: PimType) => void;
```
tasn: Could you please fix the naming here? It should be `originalTask`, or better yet, just `originalItem`.

tasn:
Oh I also just noticed Github is complaining that this branch has conflicts. Could you please take a look at that too? I think it's because I added the licensing information to files you renamed. Actually though, this reminds me, could you please make sure to include the same licensing information on the top of your newly added files too? This is what needs to be there:

```
// SPDX-FileCopyrightText: © 2017 EteSync Authors
// SPDX-License-Identifier: AGPL-3.0-only
```

AbleLincoln:
Like you said, I believe the conflicts are from moving files. GitHub doesn't allow me to resolve the conflicts. If you just accept the incoming changes it should be fine, I completely copied over those conflicting files.

tasn:
I'll resolve things locally, and merge. No problem.

tasn:
Merged, thanks!

tasn:
There are two comments I missed (no worries, I'm fixing them now), just FYI:
1. When naming function, we prefer using `function handleChange()` over `const handleChange = () => ...`
2. I think it's better to use a form element for handling the submission rather than hooking on the keypress event.
3. It was crashing when there were no task lists.
4. onItemSave was not passed.
5. Saving an item returns.

And a few other things. It's very much apparent that you haven't tested it at all. I'm quite pissed tbh for having wasted as much time as I have on this. And I'm annoyed with myself for trusting that you have tested it before pushing it. I'm reverting all of the changes. Please fix it, as it's completely broken.
I understand you took it out to its own PR so things have changed since you first wrote it, but there's no excuse to not testing! (I'm saying that to myself too!)

Additionally, please take a look at #32, the quickadd bar should be at the top, outside of the list, not a part of it... It's quite ugly this way, and much less functional with long lists...

AbleLincoln:
I was actually just coming back now to add some code I realized I forgot to push. There are just some props missing that need to be passed down.

AbleLincoln:
And I can change where the quick add bar is located, I was just going off of the design for some other apps I had researched (ios notes, Todoist).

tasn:
I edited my comment above a bit to clarify things, and tone down my frustration.

Yeah, please the position. In general, follow #32... Also, happy to answer questions if you have any. :)

tasn:
Additionally, please let me know once you'd like me to take a look.

AbleLincoln:
Ok, sorry for missing things. It's a bit difficult trying to cherry-pick out the relevant code I need for each merge. Also, design is a little wonky as I add in each individual feature. I promise everything looks better once it is all together.

tasn:
Yeah, I get, no worries. And sorry for showing my frustration, though I was just fixing one issue after the other and getting more and more annoyed until I gave up. :|

Let me know when you'd like me to take another look, though please make sure you both test and push before, and in addition to addressing the above comments.

Thanks!

tasn:
To be honest, I wouldn't return history from the `onItemSave` - it's overly specific. You should just return a Promise and if it succeeds you should go back. The history can be gotten locally with `withRouter` or can just be passed by props.

AbleLincoln:
I think I got everything, please take a look when you get a chance

src/Pim/index.tsx
```diff
@@ -107,7 +107,7 @@ type CollectionRoutesPropsType = RouteComponentProps<{}> & {
   componentEdit: any;
   componentView: any;
   items: {[key: string]: PimType};
-  onItemSave: (item: PimType, journalUid: string, originalContact?: PimType) => void;
+  onItemSave: (item: PimType, journalUid: string, originalItem?: PimType) => Promise<any>;
```
tasn: Should be Promise<void> not any.

src/Pim/index.tsx
```diff
@@ -235,11 +237,11 @@ class Pim extends React.PureComponent {
     this.onItemSave = this.onItemSave.bind(this);
   }
 
-  public onItemSave(item: PimType, journalUid: string, originalEvent?: PimType) {
+  public onItemSave(item: PimType, journalUid: string, originalEvent?: PimType): Promise<History<any>> {
```
tasn: This shouldn't return `Promise<History>` but `Promise<void>`

src/Pim/index.tsx
```diff
@@ -251,7 +253,7 @@ class Pim extends React.PureComponent {
     if (last) {
       prevUid = last.uid;
     }
-    store.dispatch<any>(fetchEntries(this.props.etesync, journal.uid, prevUid))
+    return store.dispatch<any>(fetchEntries(this.props.etesync, journal.uid, prevUid))
```
tasn: According with my comment above, this function should not early return. Just let it return like before (by getting to the end of the function). Same with the line below.

src/components/ContactEdit.tsx
```diff
@@ -113,9 +115,10 @@ interface PropsType {
   collections: EteSync.CollectionInfo[];
   initialCollection?: string;
   item?: ContactType;
-  onSave: (contact: ContactType, journalUid: string, originalContact?: ContactType) => void;
+  onSave: (contact: ContactType, journalUid: string, originalContact?: ContactType) => Promise<any>;
```
tasn: Same here with `Promise<void>` - same with all of these places, I won't comment for each.

src/components/ContactEdit.tsx
```diff
@@ -309,7 +312,10 @@ class ContactEdit extends React.PureComponent<PropsType> {
     setProperty('title', this.state.title);
     setProperty('note', this.state.note);
 
-    this.props.onSave(contact, this.state.journalUid, this.props.item);
+    this.props.onSave(contact, this.state.journalUid, this.props.item)
```
tasn: I love it. Looks much nicer! Sorry you had to deal with the mess that was there before.

src/Pim/index.tsx
AbleLincoln: `Promise<void>`?

src/Pim/index.tsx
AbleLincoln: But then how can we await the finishing of the asynchronous action?

tasn: Sorry, I thought the the function was async. Either make the function async and change the `return Promise.reject()` to `throw new Error('journalUid is undefined')` and await on these two async calls (and everything else will just work). Or you can manually create a promise. Though I think making the function async is much easier.

tasn: To be honest though, the returns you did are fine too if changing the return type to `Promise<void>` won't complain about the types.

src/Pim/index.tsx
tasn: Sorry, that's what I meant, yeah. I forgot to wrap it in code tags so the `<void>` part was stripped. It seems.

AbleLincoln: Ah yes, my favorite HTML tag: `<void>`

src/Pim/index.tsx
AbleLincoln: Yeah TypeScript doesn't seem to have a problem with the return type. `store.dispatch` is weird, it returns whatever you pass in, which in this case is of the type `Action<Promise<EteSync.Entry[]>>`. Which I guess resolves to a Promise? I'm not sure, it's a bit confusing, but it works haha.

Honestly I am ok with either way (returning promises or marking it async), it seems more like a style choice at this point. Which would you prefer?

tasn: It's typing issue which I haven't managed to solve. The return type is indeed as you said, but the actual return type is `Promise<Action<EteSync.Entry[]>>`. That's why there's an `<any>` there to workaround this issue. :|
If this works then no need to change it, it's all the same.

src/Pim/index.tsx
tasn: hehe. :laughing:

tasn:
Let me know when you'd like me to take another look at the code. Thanks!

AbleLincoln:
It's good to go

tasn:
I think you forgot to push, all of the promises are still not `Promise<void>`.

AbleLincoln:
Where? Did I miss one?

tasn:
Nah, all good. For some reason it didn't show me you pushed anything when I replied, I was still seeing the old diff.

tasn:
You added a few weird files, such as: `src/components/Tasks/TaskEdit.tsx~refs/remotes/upstream/master`.
Not sure where they came from, but please remove them.

AbleLincoln:
My bad, have no idea how that happened. Is this ok now?

tasn:
In, good job!
I had to fix one thing (you forgot to change one `onItemSave => void` to return `Promise<void>`).
I also squashed most of the commits together as we don't really want to litter the history with changes that are going back and forth between ideas. It's fine on the other hand to have multiple commits for different parts or evolution of a complex feature (step 1, 2 and etc.).

Unfortunately github doesn't detect merges when they happen from the command line, so I have to close this manually unfortunately, though the commits are in!

PR #86, opened 2020-03-12
Tasks: adds checkbox to Task list item

AbleLincoln:


tasn:
Brilliant, good job!

I have one comment that's orthogonal to this patch but relevant to the rest of your patches:
```
const TaskListItem = React.memo((props: PropsType) => {
...
});

export default TaskListItem;
```

Should be:
```
export default React.memo(function TaskListItem(props: PropsType) => {
});
```

This is not just a stylistic change, it has real implications! The difference is that the former will be listed as "Anonymous" in the react dev tools, and the latter as "TaskListItem". Without this, deugging can be quite annoying.

I don't mind doing the change myself if you prefer, but I just need to know that it won't introduce conflicts for you before I do it. So let me know.

AbleLincoln:
Oh I had no idea. That is good to know, thank you. I am new to using React memo so I'm still trying to figure out all of the intricacies.

That's totally fine to change. It won't cause any problems on my end.

tasn:
Fixed. :)

AbleLincoln:
Thank you. I'll keep this in mind moving forward.

PR #88, opened 2020-03-12
Tasks: adds priority functionality

AbleLincoln:
including Colored Radio buttons, colored checkboxes


AbleLincoln:
Yes, that is a good point. I should probably also do this conversion on the TaskListItem

src/components/Tasks/TaskEdit.tsx
```diff
@@ -88,6 +94,7 @@ class TaskEdit extends React.PureComponent<PropsType> {
       this.state.uid = event.uid;
       this.state.title = event.title ? event.title : '';
       this.state.status = event.status;
+      this.state.priority = event.priority || TaskPriorityType.Undefined;
```
tasn: This should be `??` not `||`. I know it works (because 0 happens to be falsey so it too just maps to undefined), but still better to use `??` so we put what we mean.

PR #90, opened 2020-03-13
Tasks: renames some vars from event to task

AbleLincoln:
I assume this is just some cruft from a bad copy-paste


tasn:
Nice catch. I can't believe even the user facing messages were wrong. :man_facepalming:

PR #94, opened 2020-03-14
Tasks: adds filter by tag feature

AbleLincoln:
Some of this code may look like overkill for this feature, but it is needed for upcoming features I am going to merge in.

### DIFF
--- a/src/components/Tasks/Sidebar.tsx
+++ b/src/components/Tasks/Sidebar.tsx
@@ -47,10 +47,10 @@ function SidebarListItem(props: ListItemPropsType) {
   );
 }
 
-export default function Sidebar(props: { tags: any }) {
+export default function Sidebar(props: { tags: Map<string, number> }) {
   const { tags } = props;
 
-  const tagsList = Object.entries(tags).map(([tag, amount]) => (
+  const tagsList = Array.from(tags, ([tag, amount]) => (
     <SidebarListItem
       key={tag}
       name={tag}
--- a/src/components/Tasks/Sidebar.tsx
+++ b/src/components/Tasks/Sidebar.tsx
@@ -41,7 +41,7 @@ function SidebarListItem(props: ListItemPropsType) {
 export default function Sidebar(props: { tags: Map<string, number>, totalTasks: number }) {
   const { tags, totalTasks } = props;
 
-  const tagsList = [...tags].sort(([a], [b]) => a < b ? -1 : 1).map(([tag, amount]) => (
+  const tagsList = [...tags].sort(([a], [b]) => a.localeCompare(b)).map(([tag, amount]) => (
     <SidebarListItem
       key={tag}
       name={`tag:${tag}`}
--- a/src/components/Tasks/Sidebar.tsx
+++ b/src/components/Tasks/Sidebar.tsx
@@ -55,7 +55,7 @@ function SidebarListItem(props: ListItemPropsType) {
 export default function Sidebar(props: { tags: Map<string, number> }) {
   const { tags } = props;
 
-  const tagsList = Array.from(tags, ([tag, amount]) => (
+  const tagsList = [...tags].sort(([a], [b]) => a < b ? -1 : 1).map(([tag, amount]) => (
     <SidebarListItem
       key={tag}
       name={`tag:${tag}`}
--- a/src/components/Tasks/Sidebar.tsx
+++ b/src/components/Tasks/Sidebar.tsx
@@ -14,7 +14,7 @@ import { setSettings } from '../../store/actions';
 import { StoreState } from '../../store';
 
 interface ListItemPropsType {
-  name: string;
+  name: string | null;
   Icon?: React.ComponentType<IconProps>;
   primaryText: string;
   secondaryText?: string;
@@ -61,7 +61,7 @@ export default function Sidebar(props: { tags: Map<string, number> }) {
 
   return (
     <List dense>
-      <SidebarListItem name="all" primaryText="View all tasks" />
+      <SidebarListItem name={null} primaryText="View all tasks" />
 
       <ListSubheader>Tags</ListSubheader>
       {tagsList}
--- a/src/components/Tasks/Sidebar.tsx
+++ b/src/components/Tasks/Sidebar.tsx
@@ -52,7 +52,7 @@ export default function Sidebar(props: { tags: Map<string, number> }) {
   const tagsList = Array.from(tags, ([tag, amount]) => (
     <SidebarListItem
       key={tag}
-      name={`tag ${tag}`}
+      name={`tag:${tag}`}
       primaryText={tag}
       secondaryText={String(amount)}
       Icon={LabelIcon} />
--- a/src/components/Tasks/Sidebar.tsx
+++ b/src/components/Tasks/Sidebar.tsx
@@ -38,8 +38,8 @@ function SidebarListItem(props: ListItemPropsType) {
   );
 }
 
-export default function Sidebar(props: { tags: Map<string, number> }) {
-  const { tags } = props;
+export default function Sidebar(props: { tags: Map<string, number>, totalTasks: number }) {
+  const { tags, totalTasks } = props;
 
   const tagsList = [...tags].sort(([a], [b]) => a < b ? -1 : 1).map(([tag, amount]) => (
     <SidebarListItem
@@ -53,7 +53,7 @@ export default function Sidebar(props: { tags: Map<string, number> }) {
 
   return (
     <List dense>
-      <SidebarListItem name={null} primaryText="All" icon={<InboxIcon />} />
+      <SidebarListItem name={null} primaryText="All" icon={<InboxIcon />} amount={totalTasks} />
 
       <ListSubheader>Tags</ListSubheader>
       {tagsList}
--- a/src/components/Tasks/Sidebar.tsx
+++ b/src/components/Tasks/Sidebar.tsx
@@ -2,26 +2,35 @@ import * as React from 'react';
 
 import { useSelector, useDispatch } from 'react-redux';
 
-import List from '@material-ui/core/List';
-import ListSubheader from '@material-ui/core/ListSubheader';
-import ListItem from '@material-ui/core/ListItem';
-import ListItemText from '@material-ui/core/ListItemText';
-import ListItemIcon from '@material-ui/core/ListItemIcon';
-import { IconProps } from '@material-ui/core/Icon';
+import InboxIcon from '@material-ui/icons/Inbox';
 import LabelIcon from '@material-ui/icons/LabelOutlined';
+import { makeStyles } from '@material-ui/core/styles';
 
 import { setSettings } from '../../store/actions';
 import { StoreState } from '../../store';
 
+import { List, ListItem, ListSubheader } from '../../widgets/List';
+
+const useStyles = makeStyles({
+  root: {
+    '& .MuiListItemIcon-root': {
+      minWidth: '2.5em',
+    },
+    '& .MuiListItemIcon-root:last-of-type': {
+      justifyContent: 'flex-end',
+    },
+  },
+});
+
 interface ListItemPropsType {
   name: string | null;
-  Icon?: React.ComponentType<IconProps>;
+  icon?: React.ReactElement;
   primaryText: string;
-  secondaryText?: string;
 }
 
 function SidebarListItem(props: ListItemPropsType) {
-  const { name, Icon, primaryText, secondaryText } = props;
+  const classes = useStyles();
+  const { name, icon, primaryText } = props;
   const dispatch = useDispatch();
   const taskSettings = useSelector((state: StoreState) => state.settings.taskSettings);
   const { filterBy } = taskSettings;
@@ -32,35 +41,30 @@ function SidebarListItem(props: ListItemPropsType) {
 
   return (
     <ListItem
-      button
       onClick={handleClick}
       selected={name === filterBy}
-    >
-      {Icon && (
-        <ListItemIcon>
-          <Icon fontSize="small" />
-        </ListItemIcon>
-      )}
-      <ListItemText primary={primaryText} secondary={secondaryText ?? ''} />
-    </ListItem>
+      leftIcon={icon}
+      rightIcon={<span>4</span>}
+      primaryText={primaryText}
+      customClass={classes.root}
+    />
   );
 }
 
 export default function Sidebar(props: { tags: Map<string, number> }) {
   const { tags } = props;
 
-  const tagsList = Array.from(tags, ([tag, amount]) => (
+  const tagsList = Array.from(tags, ([tag]) => (
     <SidebarListItem
       key={tag}
       name={`tag:${tag}`}
       primaryText={tag}
-      secondaryText={String(amount)}
-      Icon={LabelIcon} />
+      icon={<LabelIcon />} />
   ));
 
   return (
     <List dense>
-      <SidebarListItem name={null} primaryText="All" />
+      <SidebarListItem name={null} primaryText="All" icon={<InboxIcon />} />
 
       <ListSubheader>Tags</ListSubheader>
       {tagsList}
--- a/src/components/Tasks/Sidebar.tsx
+++ b/src/components/Tasks/Sidebar.tsx
@@ -4,23 +4,11 @@ import { useSelector, useDispatch } from 'react-redux';
 
 import InboxIcon from '@material-ui/icons/Inbox';
 import LabelIcon from '@material-ui/icons/LabelOutlined';
-import { makeStyles } from '@material-ui/core/styles';
 
 import { setSettings } from '../../store/actions';
 import { StoreState } from '../../store';
 
 import { List, ListItem, ListSubheader } from '../../widgets/List';
-
-const useStyles = makeStyles({
-  root: {
-    '& .MuiListItemIcon-root': {
-      minWidth: '2.5em',
-    },
-    '& .MuiListItemIcon-root:last-of-type': {
-      justifyContent: 'flex-end',
-    },
-  },
-});
 
 interface ListItemPropsType {
   name: string | null;
@@ -30,7 +18,6 @@ interface ListItemPropsType {
 }
 
 function SidebarListItem(props: ListItemPropsType) {
-  const classes = useStyles();
   const { name, icon, primaryText, amount } = props;
   const dispatch = useDispatch();
   const taskSettings = useSelector((state: StoreState) => state.settings.taskSettings);
@@ -45,9 +32,8 @@ function SidebarListItem(props: ListItemPropsType) {
       onClick={handleClick}
       selected={name === filterBy}
       leftIcon={icon}
-      rightIcon={<span>{amount}</span>}
+      rightIcon={<span style={{ width: '100%', textAlign: 'right' }}>{amount}</span>}
       primaryText={primaryText}
-      customClass={classes.root}
     />
   );
 }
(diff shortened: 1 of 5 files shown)


src/components/Tasks/TaskList.tsx
```diff
@@ -9,31 +9,61 @@ import * as EteSync from 'etesync';
...
 interface PropsType {
   entries: TaskType[];
   collections: EteSync.CollectionInfo[];
   onItemClick: (entry: TaskType) => void;
   onItemSave: (item: PimType, journalUid: string, originalItem?: PimType) => Promise<void>;
+  theme: Theme;
 }
 
-export default React.memo(function TaskList(props: PropsType) {
+export default React.memo(withTheme(function TaskList(props: PropsType) {
```
tasn: It's better not to use `withTheme`.
Just use `const theme = useTheme()`

src/store/actions.ts
```diff
@@ -213,7 +213,7 @@ export const clearErros = createAction(
 // FIXME: Move the rest to their own file
 export const setSettings = createAction(
   'SET_SETTINGS',
-  (settings: SettingsType) => {
+  (settings: Partial<SettingsType>) => {
```
tasn: Could you please move this and the related reducer change (just the partial part, not the task specific part) to their own commit? It's a completely unrelated change that affects the rest of the web app.

src/components/Tasks/TaskList.tsx
```diff
@@ -9,31 +9,61 @@ import * as EteSync from 'etesync';
...
+  const settings = useSelector((state: StoreState) => state.settings.taskSettings);
+  const { filterBy } = settings;
+
+  const entries = props.entries.filter((x) => showCompleted || !x.finished).filter(filters[filterBy]);
   const sortedEntries = sortSelector(entries);
 
+  // counts tags and creates an object with shape { tag: amount }
+  // TODO: memoize
+  const tags = TaskTags.reduce((obj, tag) => ({ ...obj, [tag]: 0 }), {});
+  props.entries.filter((x) => (showCompleted || !x.finished)).forEach((entry) => entry.tags.forEach((tag) => {
```
tasn: This walks through all of the tasks (including completed which could be a lot!) twice. Once above and once here.
Maybe it's a better idea to split line 57 to:
```
const potentialEntries = props.entries.filter((x) => showCompleted || !x.finished);
const entries = potentialEntries.filter(filters[filterBy]);
```
And then just change this to:
```
potentialEntries.forEach((entry) ...
```
?

src/components/Tasks/TaskList.tsx
```diff
@@ -9,31 +9,61 @@ import * as EteSync from 'etesync';
...
-  const entries = props.entries.filter((x) => showCompleted || !x.finished);
+  const settings = useSelector((state: StoreState) => state.settings.taskSettings);
+  const { filterBy } = settings;
+
+  const entries = props.entries.filter((x) => showCompleted || !x.finished).filter(filters[filterBy]);
   const sortedEntries = sortSelector(entries);
 
+  // counts tags and creates an object with shape { tag: amount }
+  // TODO: memoize
+  const tags = TaskTags.reduce((obj, tag) => ({ ...obj, [tag]: 0 }), {});
```
tasn: This is quite hacky and very slow because it creates a lot of objects. Maybe use the javascript Map?
```
const tags = new Map<string, number>();
TaskTags.forEach((tag) => tags.set(tag, tags.get(tag, 0) + 1));
```
Then you can just use this map object below using `tags.get` instead of the `hasOwnProperty` hack.

src/components/Tasks/TaskList.tsx
```diff
@@ -9,31 +9,61 @@ import * as EteSync from 'etesync';
...
 }
 
-export default React.memo(function TaskList(props: PropsType) {
+export default React.memo(withTheme(function TaskList(props: PropsType) {
   const [showCompleted, setShowCompleted] = React.useState(false);
-  const entries = props.entries.filter((x) => showCompleted || !x.finished);
+  const settings = useSelector((state: StoreState) => state.settings.taskSettings);
+  const { filterBy } = settings;
+
+  const entries = props.entries.filter((x) => showCompleted || !x.finished).filter(filters[filterBy]);
```
tasn: See below how I suggested to change it to `potentialEntries` and `entries`. If you do that, you can also do something like:
```
let entries = potentialEntries;
if (filterBy !== 'all') { // Or better yet, just have 'all' be null -> so filter by nothing.
    entries = entries.filter((x) => x.tags.includes(filterBy));
}
```

Then you can delete lines 32-43, you are more efficient in the no-filter case (the most common one) and the code is much clearer and shorter.

src/store/reducers.ts
```diff
@@ -272,13 +272,21 @@ export const errorsReducer = handleActions(
...
-      { ...action.payload }
+    [actions.setSettings.toString()]: (state: { key: string | null }, action: any) => (
+      { ...state, ...action.payload }
     ),
   },
-  { locale: 'en-gb' }
+  {
+    locale: 'en-gb',
+    taskSettings: {
+      filterBy: 'all',
```
tasn: As said elsewhere, `filterBy` should default to `null` and that should mean "filter by nothing".

src/components/Tasks/TaskList.tsx
AbleLincoln: Yes, I like that much better

src/components/Tasks/TaskList.tsx
AbleLincoln: This is much cleaner, and it will work well for filtering by tag. But I know that I also want to add the ability to filter by due date, so I can't just necessarily use `x.tags.includes(filterBy)`. That is why I have all the nonsense with creating an object of filter functions. Do you have a suggestion for a way I can handle this but still make it general enough to have any kind of filter functions?

tasn: This would have been problematic anyway, because tags called "Today" or "Recent" or whatever wouldn't have worked...
I think these are just two different concepts that should maybe be in two different fields? So the settings could be:
```
enum FilterView { ... }

filterBy: {
   tag?: string;
   filter?: FilterView
}
```
Then you can just have them as two separate `if`s that filter one after the other? (and probably have one disable the other when switching between them if you don't want them to work in tandem).

AbleLincoln: Ok, I took a slightly different approach. I didn't want to muck up `filterBy` by making it an object with multiple options. Instead, I append "tag" to any `filterBy` option that filters by tag. So for instance, if I select the "Home" tag, the `filterBy` option will be `"tag Home"`. Then I just check to see if the `filterBy` option starts with `"tag"` and I filter appropriately.

It got rid of that nasty `filters` object, and should allow things to stay pretty general.

tasn: Looks good to me. I'm going to take a look at the whole patch again now.

AbleLincoln: Awesome, thank you

src/components/Tasks/Sidebar.tsx
```diff
@@ -0,0 +1,70 @@
...
+      onClick={handleClick}
+      selected={name === filterBy}
+    >
+      {Icon ?
+        <ListItemIcon>
+          <Icon fontSize="small" />
+        </ListItemIcon> :
+        ''
+      }
+      <ListItemText primary={primaryText} secondary={secondaryText || ''} />
```
tasn: Should be `?? ''`

src/components/Tasks/Sidebar.tsx
```diff
@@ -0,0 +1,70 @@
...
+  );
+}
+
+export default function Sidebar(props: { tags: Map<string, number> }) {
+  const { tags } = props;
+
+  const tagsList = Array.from(tags, ([tag, amount]) => (
+    <SidebarListItem
+      key={tag}
+      name={`tag ${tag}`}
```
tasn: Horrible bike-shedding, I know, but could you please change this to `tag:${tag}`? So have a colon as the delimiter? I know it's just my personal preference, but using space as the delimiter fields very weird for me. I was trying to not mention it (and I wasn't going it if I had no other comments), but since I'm already commenting... :)

src/components/Tasks/TaskList.tsx
```diff
@@ -31,9 +38,26 @@ interface PropsType {
...
+  const { filterBy } = settings;
+  const theme = useTheme();
+
+  const potentialEntries = props.entries.filter((x) => showCompleted || !x.finished);
+  let entries;
+  if (filterBy?.startsWith('tag')) {
+    const tag = filterBy.slice(4);
+    entries = potentialEntries.filter((x) => x.tags.includes(tag));
+  } else {
+    entries = [...potentialEntries];
```
tasn: Why are you cloning the list? Just use the potentialEntries list directly. Cloning for no reason is inefficient. :)

src/components/Tasks/TaskList.tsx
```diff
@@ -31,9 +38,26 @@ interface PropsType {
...
   const [showCompleted, setShowCompleted] = React.useState(false);
-  const entries = props.entries.filter((x) => showCompleted || !x.finished);
+  const settings = useSelector((state: StoreState) => state.settings.taskSettings);
+  const { filterBy } = settings;
+  const theme = useTheme();
+
+  const potentialEntries = props.entries.filter((x) => showCompleted || !x.finished);
+  let entries;
+  if (filterBy?.startsWith('tag')) {
+    const tag = filterBy.slice(4);
```
tasn: In continuation to my comment about making the delimiter `:`, the above should probably be changed to `startsWith('tag:')` (now there's a bug if we ever add another word that starts with tag).

Having the magic number 4 here feels a bit wrong to me tbh. `filterBy.slice(filterBy.indexOf(':') + 1)` is ugly as hell, though maybe better? Or maybe even:
```
const tagPrefix = 'tag:';
if (filterBy?.startsWith(tagPrefix)) {
    const tag = filterBy.slice(tagPrefix.length);
```

src/components/Tasks/Sidebar.tsx
```diff
@@ -0,0 +1,70 @@
...
+      primaryText={tag}
+      secondaryText={String(amount)}
+      Icon={LabelIcon} />
+  ));
+
+  return (
+    <List dense>
+      <SidebarListItem name={null} primaryText="View all tasks" />
```
tasn: I think "My Tasks" would be a better name, or maybe simply "All"?
It looks odd (and overly verbose) having this as an imperative (View...) and have the word tasks, and the rest as just tags.

src/components/Tasks/Sidebar.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+export default function Sidebar(props: { tags: Map<string, number> }) {
+  const { tags } = props;
+
+  const tagsList = Array.from(tags, ([tag, amount]) => (
+    <SidebarListItem
+      key={tag}
+      name={`tag ${tag}`}
+      primaryText={tag}
+      secondaryText={String(amount)}
```
tasn: I think it's better to just pass the numerical amount to the `SidebarListItem` and then let it decide how to design itself. Here you are kind of forcing it from the outside to put it in the secondaryText field which feels off.

The reason why I even noticed this forcing from the outside, is because I think the count looks quite odd alone there on its own line. Maybe have it as a pill (badge?) next to the name? So it would look something like: `Work [2]` where the square brackets represent a pill/badge?
Or maybe follow the example in #32 and just have it aligned to the right (and make the sidebar larger)?

src/components/Tasks/Sidebar.tsx
```diff
@@ -0,0 +1,70 @@
+import * as React from 'react';
+
+import { useSelector, useDispatch } from 'react-redux';
+
+import List from '@material-ui/core/List';
```
tasn: We have our own wrapper for List in `widgets`. If it's not too much work, it would be great if you could change to use that for more consistency in the code (we use it everywhere else too). No worries if it's a lot of hassle.

AbleLincoln: That's fine, I can change it to that.

I was actually curious, why are we using this custom List component instead of the built in Material UI one?

src/components/Tasks/Sidebar.tsx
AbleLincoln: That makes sense. I was also debating whether I should use a colon or something for the delimiter.

src/components/Tasks/Sidebar.tsx
tasn: As you can see by the code you had to write to use it there are a lot of components involved when creating list items, which in turn meant abysmal performance when using it in large lists. This was even worse with earlier versions of material-ui.
In order to solve it we just implemented our own list widget that looked the same and was orders of magnitude faster.

As of a few days ago (5cb8dda9cce94abe4aa57735548960706bce61c1) we changed back to use the material-ui widgets internally because we finally got around to adding virtualized lists, which fixes the performance problem altogether.

src/components/Tasks/TaskList.tsx
AbleLincoln: This was actually intentional. There was some weirdness with `entries` being a reference to `potentialEntries`. Since I was sorting `entries`, it was also sorting `potentialEntries`. And since the tags list used `potentialEntries`, the order of the tags would unexpectedly change when selecting different filtering options.

I found a dumb simple solution to this though: just sort the entries _after_ getting the tags.

src/components/Tasks/Sidebar.tsx
AbleLincoln: Ahh that makes sense. That mui library is quite bulky haha

AbleLincoln: Is it ok if I change the widget? There is no support for `selected` yet

src/components/Tasks/TaskList.tsx
tasn: You should probably filter the tags alphabetically anyway, so I guess it doesn't matter if you sort before or after.

src/components/Tasks/Sidebar.tsx
tasn: Sure thing.

src/components/Tasks/Sidebar.tsx
AbleLincoln: I've changed the design to match #32 more closely. I had to do some finagling with passing styles down to the `ListItem` component. I'm not sure if I did this correctly, or if this is the best way to do it, so please let me know your thoughts.

src/components/Tasks/Sidebar.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { makeStyles } from '@material-ui/core/styles';
+
+import { setSettings } from '../../store/actions';
+import { StoreState } from '../../store';
+
+import { List, ListItem, ListSubheader } from '../../widgets/List';
+
+const useStyles = makeStyles({
+  root: {
+    '& .MuiListItemIcon-root': {
```
tasn: Why did you have to do that? Rely on internal class names like this is a bad idea and a sure way to get breaks down the line.
You can just use the `rightIcon` property and put whatever you want there, e.g:
`rightIcon={<Badge>{count}</Badge}` or even `rightIcon={<span>{count}</span>}`.

Edit: I see you already did it excatly this way, so why the extra classes? What's failing? Maybe it's something we can fix in the underlying code.

src/components/Tasks/Sidebar.tsx
```diff
@@ -0,0 +1,76 @@
...
+      primaryText={primaryText}
+      customClass={classes.root}
+    />
+  );
+}
+
+export default function Sidebar(props: { tags: Map<string, number> }) {
+  const { tags } = props;
+
+  const tagsList = Array.from(tags, ([tag, amount]) => (
```
tasn: It looks like you forgot to sort the tags.

src/widgets/List.tsx
```diff
@@ -45,6 +45,8 @@ interface ListItemPropsType {
   href?: string;
   insetChildren?: boolean;
   nestedItems?: React.ReactNode[];
+  selected?: boolean;
+  customClass?: any;
```
tasn: Why not call it `className` and propagate that? (including having the correct type for it)? It's essentially what you are doing just with a non-standard name.

src/components/Tasks/Sidebar.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const handleClick = () => {
+    dispatch(setSettings({ taskSettings: { ...taskSettings, filterBy: name } }));
+  };
+
+  return (
+    <ListItem
+      onClick={handleClick}
+      selected={name === filterBy}
+      leftIcon={icon}
+      rightIcon={<span>{amount}</span>}
```
tasn: Maybe use a chip rather than a span? Whatever you think looks better, though if you are using a chip, use a sort of grey one, not a flashy blue one.

src/components/Tasks/Sidebar.tsx
tasn: I think it's a bit hacky, and commented there.

src/components/Tasks/Sidebar.tsx
AbleLincoln: it's to deal with these ugly styles. This is how it looks without my custom styles. See how the number just floats there? I want to right-align it some way.
Screen Shot 2020-03-17 at 2 55 42 AM

I think part of the reason I have to deal with it like this is that the MUI ListItem API doesn't really provide a way to have some secondary, right-aligned icon. The closest is `ListItemSecondaryAction`, but using that seems semantically incorrect.

Could we modify our `ListItem` to accommodate right-aligning the `rightIcon`?

tasn: You just need to set the following on the span:
`<span style={{ width: '100%', textAlign: 'right' }}>{amount}</span>` and it should work. No need to hack it from the outside. :)

AbleLincoln: Nice, that is much simpler than what I had

AbleLincoln: It's not a huge deal, but I was also trying to reduce the space between the left icon and the text. Just seems a little too wide to me. Do you have any suggestions for how I could handle that?

src/components/Tasks/Sidebar.tsx
```diff
@@ -0,0 +1,62 @@
...
+      rightIcon={<span style={{ width: '100%', textAlign: 'right' }}>{amount}</span>}
+      primaryText={primaryText}
+    />
+  );
+}
+
+export default function Sidebar(props: { tags: Map<string, number>, totalTasks: number }) {
+  const { tags, totalTasks } = props;
+
+  const tagsList = [...tags].sort(([a], [b]) => a < b ? -1 : 1).map(([tag, amount]) => (
```
tasn: The common way to do string comparison is `a.localeCompare(b)`.

src/components/Tasks/Sidebar.tsx
tasn: I wouldn't touch that. It's part of the theme and how the library does lists.

AbleLincoln: Ok I'll just leave it as it

tasn:
Hey,

I merged (and squashed) this patch. I also changed the first commit a bit because it wasn't split correctly (I figured it would be faster to just do it myself rather than another back and forth).
The idea of having the settings commit separate was because you changed something fundamental in how the settings action works which is unrelated to the rest of the patch. So you mixing that change with the "adding tasks settings" kinda ruined the point of separating them. :)

Anyhow, it's in! \o/

tasn:
OMG, I'm so sorry! I just realised my squash messed up with the authorship!!! :(

tasn:
Fixed. :)

PR #96, opened 2020-03-17
Tasks: makes tags clickable to trigger filter

AbleLincoln:
You can click on the individual tags on each ListItem and it will filter tasks by that tag.

This might be a silly feature, I just had the idea and wanted to test it out. It doesn't need to be merged in if it's too clunky.

### DIFF
--- a/src/components/Tasks/TaskListItem.tsx
+++ b/src/components/Tasks/TaskListItem.tsx
@@ -15,6 +15,10 @@ import moment from 'moment';
 
 import { mapPriority } from '../../helpers';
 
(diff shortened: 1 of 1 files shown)


tasn:
Ah, interesting idea!

I'm on the the fence about it to be honest. On one hand, it may be a nice shortcut, on the other, there's already a button for it on the sidebar. I'm also concerned about it being annoying on mobile where the touch area is already limited.

AbleLincoln:
Yeah that would be frustrating. Also, it's not really clear that they are clickable. No worries

PR #97, opened 2020-03-18
Tasks: adds memoization to Sidebar

AbleLincoln:


tasn:
Great job, and much more elegant without specific memoization in the sidebar, but instead just having it rely on the props not changing and React.memo.

AbleLincoln:
Thank you 😄 makes the Sidebar render waaaaay less often

tasn:
I don't know if it's just me, but things feel faster indeed. :)

PR #98, opened 2020-03-18
Tasks: adds dueToday getter to TaskType

AbleLincoln:
cleans up some code around filtering by due today


tasn:
Much nicer indeed.

PR #99, opened 2020-03-18
Tasks: moves show completed toggle to sidebar

AbleLincoln:
Changes input to a toggle.
Also, `showCompleted` is in redux store

### DIFF
--- a/src/components/Tasks/Sidebar.tsx
+++ b/src/components/Tasks/Sidebar.tsx
@@ -5,8 +5,6 @@ import { useSelector, useDispatch } from 'react-redux';
 import InboxIcon from '@material-ui/icons/Inbox';
 import LabelIcon from '@material-ui/icons/LabelOutlined';
 import TodayIcon from '@material-ui/icons/Today';
-import Switch from '@material-ui/core/Switch';
-import Divider from '@material-ui/core/Divider';
 
 import { setSettings } from '../../store/actions';
 import { StoreState } from '../../store';
@@ -42,23 +40,6 @@ function SidebarListItem(props: ListItemPropsType) {
   );
 }
 
-function ToggleCompleted() {
-  const dispatch = useDispatch();
-  const taskSettings = useSelector((state: StoreState) => state.settings.taskSettings);
-  const { showCompleted } = taskSettings;
-
-  const handleChange = (_e: React.ChangeEvent<HTMLInputElement>, checked: boolean) => {
-    dispatch(setSettings({ taskSettings: { ...taskSettings, showCompleted: checked } }));
-  };
-
-  return (
-    <ListItem
-      primaryText="Completed"
-      leftIcon={<Switch edge="start" checked={showCompleted} onChange={handleChange} />}
-    />
-  );
-}
-
 export default React.memo(function Sidebar(props: { tasks: TaskType[] }) {
   const { tasks } = props;
 
@@ -80,16 +61,12 @@ export default React.memo(function Sidebar(props: { tasks: TaskType[] }) {
   ));
 
   return (
-    <List>
+    <List dense>
       <SidebarListItem name={null} primaryText="All" icon={<InboxIcon />} amount={tasks.length} />
       <SidebarListItem name="today" primaryText="Due today" icon={<TodayIcon />} amount={amountDueToday} />
 
       <ListSubheader>Tags</ListSubheader>
       {tagsList}
-
-      <Divider style={{ marginTop: '1em', marginBottom: '1em' }} />
-
-      <ToggleCompleted />
     </List>
   );
 });
--- a/src/components/Tasks/Sidebar.tsx
+++ b/src/components/Tasks/Sidebar.tsx
@@ -5,6 +5,8 @@ import { useSelector, useDispatch } from 'react-redux';
 import InboxIcon from '@material-ui/icons/Inbox';
 import LabelIcon from '@material-ui/icons/LabelOutlined';
 import TodayIcon from '@material-ui/icons/Today';
+import Switch from '@material-ui/core/Switch';
+import Divider from '@material-ui/core/Divider';
 
 import { setSettings } from '../../store/actions';
 import { StoreState } from '../../store';
@@ -40,6 +42,23 @@ function SidebarListItem(props: ListItemPropsType) {
   );
 }
 
+function ToggleCompleted() {
+  const dispatch = useDispatch();
+  const taskSettings = useSelector((state: StoreState) => state.settings.taskSettings);
+  const { showCompleted } = taskSettings;
+
+  const handleChange = (_e: React.ChangeEvent<HTMLInputElement>, checked: boolean) => {
+    dispatch(setSettings({ taskSettings: { ...taskSettings, showCompleted: checked } }));
+  };
+
+  return (
+    <ListItem
+      primaryText="Completed"
+      leftIcon={<Switch edge="start" checked={showCompleted} onChange={handleChange} />}
+    />
+  );
+}
+
 export default React.memo(function Sidebar(props: { tasks: TaskType[] }) {
   const { tasks } = props;
 
@@ -61,12 +80,16 @@ export default React.memo(function Sidebar(props: { tasks: TaskType[] }) {
   ));
 
   return (
-    <List dense>
+    <List>
       <SidebarListItem name={null} primaryText="All" icon={<InboxIcon />} amount={tasks.length} />
       <SidebarListItem name="today" primaryText="Due today" icon={<TodayIcon />} amount={amountDueToday} />
 
       <ListSubheader>Tags</ListSubheader>
       {tagsList}
+
+      <Divider style={{ marginTop: '1em', marginBottom: '1em' }} />
+
+      <ToggleCompleted />
     </List>
   );
 });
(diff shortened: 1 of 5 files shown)


src/components/Tasks/Sidebar.tsx
```diff
@@ -61,12 +80,16 @@ export default React.memo(function Sidebar(props: { tasks: TaskType[] }) {
...
   return (
-    <List dense>
+    <List>
       <SidebarListItem name={null} primaryText="All" icon={<InboxIcon />} amount={tasks.length} />
       <SidebarListItem name="today" primaryText="Due today" icon={<TodayIcon />} amount={amountDueToday} />
 
       <ListSubheader>Tags</ListSubheader>
       {tagsList}
+
+      <Divider style={{ marginTop: '1em', marginBottom: '1em' }} />
```
tasn: You can just use 'marginVertical'

AbleLincoln: oh cool I had no idea. Is that a React feature? or MUI?

tasn: Please ignore me, it's a react-native thing, not a reactjs thing.

AbleLincoln:
I liked it better there too, but I'm kind of running out of space. I have to fit sorting, searching, and these toggles somewhere. When I tried to place the sorting on the top where the toggle used to be, it looked weird. That's why I am moving the toggle in the first place.

That screen shot is great, and I've tried to refer to it. Our design is just a little different with the nav tabs and paper and everything. I wonder, do you think I should make some kind of toolbar like the example? Or I could do a three dot menu, like on Tasks.org?

AbleLincoln:
Also, I see what you're saying about settings and pages. My main reason I was using the settings object was because I wanted some kind of persistent state. From a UX perspective it's frustrating that every time I go back Tasks list (whether from a different page, or from editing) it's in some default state, rather than the way I left it.

tasn:
> I liked it better there too, but I'm kind of running out of space. I have to fit sorting, searching, and these toggles somewhere. When I tried to place the sorting on the top where the toggle used to be, it looked weird. That's why I am moving the toggle in the first place.
> 
> That screen shot is great, and I've tried to refer to it. Our design is just a little different with the nav tabs and paper and everything. I wonder, do you think I should make some kind of toolbar like the example? Or I could do a three dot menu, like on Tasks.org?

I'd say: very similar to Tasks.org. A search icon directly there, maybe also sort (though probably not) and then the rest in the dot menu. Sort, show completed and etc aren't *that* common to normal workflow.

tasn:
> Also, I see what you're saying about settings and pages. My main reason I was using the settings object was because I wanted some kind of persistent state. From a UX perspective it's frustrating that every time I go back Tasks list (whether from a different page, or from editing) it's in some default state, rather than the way I left it.

You can also have it in the navigation context, or even the navigation path itself. I don't know what's better so don't worry about it. Maybe sometime in the future I (or you) will change it. I just wanted to share my thoughts on this in case it affects your future work on this.

As for the "every time I go back", it should work just fine when you use the navigation context, that's actually what it's for. Think of it as kind of a query string, so the browser (and the app) already know how to deal with it.

AbleLincoln:
Try this. I've moved the option to a drop down menu

AbleLincoln:
> You can also have it in the navigation context, or even the navigation path itself. 

Interesting, then the "state" is controlled by the router essentially?

I'll keep this in mind as I develop, and we discuss refactoring to this after the feature is more complete.

tasn:
It's "state" in a similar way to how to how the url is state (though admittedly a bit more). Yeah, though, the nice thing about it is that it maintains the stackable nature of state.

src/components/Tasks/Sidebar.tsx
```diff
@@ -61,7 +61,7 @@ export default React.memo(function Sidebar(props: { tasks: TaskType[] }) {
   ));
 
   return (
-    <List dense>
```
tasn: This is an unrelated change to this PR.

src/components/Tasks/TaskList.tsx
```diff
@@ -37,10 +35,10 @@ interface PropsType {
 }
 
 export default function TaskList(props: PropsType) {
-  const [showCompleted, setShowCompleted] = React.useState(false);
   const settings = useSelector((state: StoreState) => state.settings.taskSettings);
   const { filterBy } = settings;
   const theme = useTheme();
+  const [showCompleted, setShowCompleted] = React.useState(false);
```
tasn: Please move it back to here it was (look at the diff of the change as a whole). It's just an unrelated change now.

src/components/Tasks/Toolbar.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+import QuickAdd from './QuickAdd';
+
+import { PimType } from '../../pim-types';
+
+interface PropsType {
+  defaultCollection: EteSync.CollectionInfo;
+  onItemSave: (item: PimType, journalUid: string, originalItem?: PimType) => Promise<void>;
+  showCompleted: boolean;
+  setShowCompleted: React.Dispatch<React.SetStateAction<boolean>>;
```
tasn: This is not the correct type (though it seems to work for you). The correct type is `(completed: boolean) => void`.

src/components/Tasks/Toolbar.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export default function Toolbar(props: PropsType) {
+  const { defaultCollection, onItemSave, showCompleted, setShowCompleted } = props;
+
+  const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
```
tasn: You are probably looking for https://reactjs.org/docs/refs-and-the-dom.html#creating-refs rather than manually setting the ref like this. Though I see that's the recommended way for MUI, so I guess keep it this way. Though :nauseated_face:

src/components/Tasks/Toolbar.tsx
```diff
@@ -0,0 +1,67 @@
...
+          aria-haspopup="true"
+          onClick={handleClick}
+        >
+          <MoreVertIcon />
+        </IconButton>
+        <Menu
+          id="simple-menu"
+          anchorEl={anchorEl}
+          keepMounted
+          open={Boolean(anchorEl)}
```
tasn: We use `!!` for boolean coercion in the rest of the code, so if you can change here too.

src/components/Tasks/Toolbar.tsx
```diff
@@ -0,0 +1,67 @@
...
+    setAnchorEl(event.currentTarget);
+  };
+
+  const handleClose = () => {
+    setAnchorEl(null);
+  };
+
+  return (
+    <div style={{ display: 'flex', justifyContent: 'space-between' }}>
+      {defaultCollection && <QuickAdd onSubmit={onItemSave} defaultCollection={defaultCollection} />}
```
tasn: I missed it in a previous change, but it looks like there's a hardcoded `25px` right margin in QuickAdd. It should be removed from there entirely (as this is not an internal property of QuickAdd) and instead added here.
The spacing is also wrong because:
1. It's too spacious.
2. It's using pixel sizing rather than `em`.
It should be changed to `0.75em` which I checked and looks good locally.

src/components/Tasks/Toolbar.tsx
```diff
@@ -0,0 +1,67 @@
...
+          id="simple-menu"
+          anchorEl={anchorEl}
+          keepMounted
+          open={Boolean(anchorEl)}
+          onClose={handleClose}
+        >
+          <MenuItem>
+            <FormControlLabel
+              label="Show completed"
+              control={<Switch checked={showCompleted} onChange={(_e, checked) => setShowCompleted(checked)} />}
```
tasn: I think that if it's a switch, the switch should be on the left and the label on the right, and if it's checkbox the way it is now. What do you think? It just looks odd to me like this.

src/components/Tasks/Toolbar.tsx
AbleLincoln: Yeah I just took it directly off their docs 😬

src/components/Tasks/Toolbar.tsx
AbleLincoln: Do you mean the other way around? I think the switch is currently on the left

tasn: Yeah, the other way around. Sorry, I meant switch on the right. :)

AbleLincoln: I put it that way and you're right, looks much better.

tasn: Great. :)

src/components/Tasks/Toolbar.tsx
```diff
@@ -0,0 +1,68 @@
...
+    setAnchorEl(event.currentTarget);
+  };
+
+  const handleClose = () => {
+    setAnchorEl(null);
+  };
+
+  return (
+    <div style={{ display: 'flex', justifyContent: 'space-between' }}>
+      {defaultCollection && <QuickAdd style={{ flexGrow: 1, marginRight: '0.75em' }} onSubmit={onItemSave} defaultCollection={defaultCollection} />}
```
tasn: Ah, nice catch moving the flexGrow too, I missed that one. :)

PR #100, opened 2020-03-18
Tasks: changes dueToday to include all times before today

AbleLincoln:


tasn:
Damn, momentjs is handy. :)

AbleLincoln:
Haha yeah that was probably the shortest change I've ever had to make

PR #101, opened 2020-03-18
Tasks: hides Sidebar amount if value 0

AbleLincoln:


src/components/Tasks/Sidebar.tsx
```diff
@@ -34,7 +34,7 @@ function SidebarListItem(props: ListItemPropsType) {
       onClick={handleClick}
       selected={name === filterBy}
       leftIcon={icon}
-      rightIcon={<span style={{ width: '100%', textAlign: 'right' }}>{amount}</span>}
+      rightIcon={<span style={{ width: '100%', textAlign: 'right' }}>{amount > 0 ? amount : ''}</span>}
```
tasn: I don't know why eslint didn't error on this, but we always put parenthesis around the condition in the conditional ternary operation. So `(amount > 0) ? amount : ''`. Though here it should probably just be `(amount > 0) && amount`

PR #104, opened 2020-03-20
Tasks: adds sorting

AbleLincoln:


src/store/reducers.ts
```diff
@@ -274,6 +274,7 @@ export interface SettingsType {
   locale: string;
   taskSettings: {
     filterBy: string | null;
+    sortBy: string;
```
tasn: Please add a default value for it (a few lines above).

In addition, please add a settings migration, so it's always set. Look at `cacheMigrations` in `src/store/construct.ts`.

src/components/Tasks/Toolbar.tsx
```diff
@@ -35,7 +50,19 @@ export default function Toolbar(props: PropsType) {
 
   return (
     <div style={{ display: 'flex', justifyContent: 'space-between' }}>
-      {defaultCollection && <QuickAdd style={{ flexGrow: 1, marginRight: '0.75em' }} onSubmit={onItemSave} defaultCollection={defaultCollection} />}
+      {defaultCollection && <QuickAdd style={{ flexGrow: 1, marginRight: '1.75em' }} onSubmit={onItemSave} defaultCollection={defaultCollection} />}
+
+      <FormControl style={{ marginRight: '0.75em' }}>
+        <InputLabel>Sort by</InputLabel>
```
tasn: This looks terrible. It shouldn't be an `InputLabel`... Just make it a sort icon that opens the menu...
After you do that you will probably want to change the top margin right accordingly.
And change the menu item that is currently selected to indicate it in some way.

src/components/Tasks/TaskList.tsx
```diff
@@ -57,7 +115,7 @@ export default function TaskList(props: PropsType) {
     entries = potentialEntries;
   }
 
-  const sortedEntries = sortSelector(entries);
+  const sortedEntries = entries.sort(getSortFunction(sortBy ?? 'smart'));
```
tasn: Why no selector here? You will be sorting on every render.

src/components/Tasks/TaskList.tsx
```diff
@@ -22,10 +20,70 @@ import Toolbar from './Toolbar';
...
+}
+
+function sortDueDate(aIn: TaskType, bIn: TaskType) {
+  const a = aIn.dueDate?.toJSDate() ?? new Date(0);
+  const b = bIn.dueDate?.toJSDate() ?? new Date(0);
+  return (a > b) ? -1 : (a < b) ? 1 : 0;
+}
+
+function sortPriority(aIn: TaskType, bIn: TaskType) {
+  const a = aIn.priority || Infinity;
```
tasn: Are you intentionally turning zero here to infinity?

src/components/Tasks/TaskList.tsx
tasn: Also the type of `sortBy` doesn't allow for null or undefined, so you shouldn't have this test here. The reason why you need it is because you didn't add state migration and a default value. Once you fix that, this can be removed.

src/store/reducers.ts
AbleLincoln: Do I need to migrate values that don't matter, like `locale`?

tasn: Nah. Take a look at the example I mentioned: https://github.com/etesync/etesync-web/blob/master/src/store/construct.ts#L154
You just need to do something like:
```
return {
  ...state,
  taskSettings: {
    ...state.taskSettings,
    sortBy: 'smart',
  },
};
```

src/components/Tasks/TaskList.tsx
AbleLincoln: I figured anything that would cause a re-render (toggle show completed, filter, sort, adding/editing a task) would also require a sort anyway, so using a selector seemed kind of pointless

tasn: I have no idea, doesn't a navigation cause a re-render? We saw it in a different PR. I'm not arguing one way or the other, it's just something that popped.

src/components/Tasks/TaskList.tsx
AbleLincoln: Yes. The way priorities work is kind of dumb. 1 is the highest priority and 9 is the lowest, but 0 is undefined. The comparator math doesn't work out unless you make undefined the highest amount (else it sorts undefined priorities to the top).

src/components/Tasks/TaskList.tsx
AbleLincoln: Although it might be worth memoizing `getSortFunction`? 🤔

src/components/Tasks/TaskList.tsx
tasn: Maybe add a comment before this line then? I just feel like it's one of those things that someone down the line will see and think it was a mistake and change it to `??` (which is wrong).

src/components/Tasks/TaskList.tsx
tasn: Probably not... It's a very short function. The sorting can get massive if you have many tasks (especially when showCompleted is set).

src/store/reducers.ts
AbleLincoln: Just committed, let me know if I did it right.

tasn: Looks right to me, and I also assume it works?

src/components/Tasks/TaskList.tsx
AbleLincoln: > doesn't a navigation cause a re-render

Ah I guess that is true. Still though, I think the number of cases where the selector would add efficiency is slim. I think that's partly due to the way we currently handle sorting and filtering. Right now the entries are first filtered then sorted, so the "entries" that are passed to the sorter are often different on each render (unless there is no filter option selected).

We could memoize more of this whole process. Or sort before filtering, in order to take advantage of `reselect` more.

AbleLincoln: I also have a question: why were we using the `reselect` library? Isn't that more for state? Does it differ at all from `useMemo`?

src/store/reducers.ts
AbleLincoln: Yep, works great!

src/components/Tasks/TaskList.tsx
tasn: The filtering is also memoized no? So you would get the same thing...
Anyhow, it doesn't matter, don't worry about it now, we can just do it in the future if we see any performance issues.

tasn: > I also have a question: why were we using the `reselect` library? Isn't that more for state? Does it differ at all from `useMemo`?

We initially used `reselect` because `useMemo` didn't exist when we first wrote this code. :)
Reselect has a few small advantages over `useMemo` that you can use it outside of function components, you can use it conditionally, and you can share the same memoization across components.

tasn:
Anyhow, let me know once you finished everything and you would like me to review it again.

src/components/Tasks/TaskList.tsx
AbleLincoln: The filtering is not memoized yet, but it probably should be. Like you said, I'll put it in my TODO list for performance improvements and we can reassess then.

AbleLincoln: Oh ok cool. This might be a good use case to take advantage of `reselect`'s composition features too

AbleLincoln:
I think that's everything. Please review when you have a chance

src/components/Tasks/Toolbar.tsx
```diff
@@ -23,16 +29,21 @@ interface PropsType {
...
-    setAnchorEl(event.currentTarget);
-  };
+  const dispatch = useDispatch();
+  const taskSettings = useSelector((state: StoreState) => state.settings.taskSettings);
+  const { sortBy } = taskSettings;
 
-  const handleClose = () => {
-    setAnchorEl(null);
+  const handleSortChange = (sort: string) => {
+    dispatch(setSettings({ taskSettings: { ...taskSettings, sortBy: sort } }));
```
tasn: Please also close the menu here (only for the sorting, not for the options menu).

tasn:
Oh, I also just realised that dueDate is being sorted in reverse... Could you please also fix that? Older dueDates should be more to the top.

AbleLincoln:
> Oh, I also just realised that dueDate is being sorted in reverse... Could you please also fix that? Older dueDates should be more to the top (and no dueDate, like now, last).

I handled this in kind of a weird way in order to keep the code succinct. Is this ok?

src/components/Tasks/Toolbar.tsx
```diff
@@ -23,16 +29,24 @@ interface PropsType {
...
+    setSortAnchorEl(null);
   };
 
+  const SortMenuItem = React.forwardRef(function SortMenuItem(props: { name: string, label: string }, _ref) {
```
tasn: I'm a bit confused, why this ref forwarding? It doesn't look needed, and I don't know why you added it.

tasn:
> > Oh, I also just realised that dueDate is being sorted in reverse... Could you please also fix that? Older dueDates should be more to the top (and no dueDate, like now, last).
> 
> I handled this in kind of a weird way in order to keep the code succinct. Is this ok?

Looks fine to me.

src/components/Tasks/Toolbar.tsx
AbleLincoln: There was an error in the console that I had missed: `Warning: Function components cannot be given refs`. I guess the `Menu` component sets a ref on its `MenuItem` children (see [MUI source code](https://github.com/mui-org/material-ui/blob/1159c65e31a6b992b7eda42fb63ccaa9aee5a946/packages/material-ui/src/Menu/Menu.js#L118-L127))

tasn: I assume then you should forward the ref to the `MenuItem` and not just ignore it.

AbleLincoln: yes, haha, just realized that

AbleLincoln: fixed

tasn: Grr, I just realised we have this warning in the side menu too (unrelated to your changes). I'll have to take a look later. :|

PR #105, opened 2020-03-23
Tasks: improves positioning of Toolbar menus

AbleLincoln:
Menus previously blocked menu buttons

### DIFF
--- a/src/widgets/Menu.tsx
+++ b/src/widgets/Menu.tsx
@@ -19,10 +19,10 @@ const transferOrigin: PopoverOrigin = {
 export default function Menu(props: MenuProps) {
   return (
     <MuiMenu
-      {...props}
       getContentAnchorEl={null}
       anchorOrigin={anchorOrigin}
       transformOrigin={transferOrigin}
+      {...props}
     />
   );
 }
--- a/src/widgets/Menu.tsx
+++ b/src/widgets/Menu.tsx
@@ -1,0 +1,43 @@
+// SPDX-FileCopyrightText: © 2017 EteSync Authors
+// SPDX-License-Identifier: AGPL-3.0-only
+
+import * as React from 'react';
+
+import MuiMenu from '@material-ui/core/Menu';
+import { PopoverOrigin } from '@material-ui/core/Popover';
+
+const anchorOrigin: PopoverOrigin = {
+  vertical: 'bottom',
+  horizontal: 'right',
+};
+
+const transferOrigin: PopoverOrigin = {
+  vertical: 'top',
+  horizontal: 'right',
+};
+
+interface PropsType {
+  anchorEl?: Element | ((element: Element) => Element) | null | undefined;
+  keepMounted?: boolean;
+  open: boolean;
+  onClose: ((event: {}, reason: 'backdropClick' | 'escapeKeyDown') => void);
+  children?: React.ReactNode | React.ReactNode[];
+}
+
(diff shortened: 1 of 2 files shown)


tasn:
That's how things behave on Android, and I don't have a problem with that... Is there a specific reason why you are changing it?

AbleLincoln:
Menus obscuring the buttons felt strange to me, so I was looked at how some other similar web apps were doing things (especially Google products). They all seem to open the menu below the button. For example, here's Google Keep. I think on the web this makes more sense (because space isn't so limited). It also allows the user to close the menu by clicking on the button again.

Screen Shot 2020-03-23 at 12 50 20 AM

tasn:
Interesting. As I said, I have no objection, I was just wondering where it was coming from...

Anyhow, if you want to do it (and assuming this is the right way, haven't checked that yet), it's probably better to create a widget under `widgets/` and change all of the usage of menu to use this, so the app is consistent (I don't think there are many/any others, but better to make sure).

AbleLincoln:
How does this look? I've moved the Menu to a widget. I couldn't find any other uses of Menu in the app. There are some `MenuItems` in use with `Selects`, but I think these are different UI so I just left them alone.

tasn:
Could you maybe just have `Menu` have the same props type as `MuiMenu` and just pass them all directly to the inner Menu? This will save us from needing to manually add the props (especially since you probably haven't covered all of them).

AbleLincoln:
Is this more generic?

src/widgets/Menu.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+const transferOrigin: PopoverOrigin = {
+  vertical: 'top',
+  horizontal: 'right',
+};
+
+export default function Menu(props: MenuProps) {
+  return (
+    <MuiMenu
+      {...props}
```
tasn: Move the spread after the rest of the props so that people can pass props to override the default ones.

src/widgets/Menu.tsx
```diff
@@ -0,0 +1,28 @@
...
+      getContentAnchorEl={null}
```
tasn: What's this, and why is it needed? I'd expect it to do something fancy like checking the menu fits the screen in this specific orientation. Make sure that's not broken.

AbleLincoln: It's to surpress this error:
```
you can not change the default `anchorOrigin.vertical` value 
when also providing the `getContentAnchorEl` prop to the popover component.
Only use one of the two props.
Set `getContentAnchorEl` to `null | undefined` or leave `anchorOrigin.vertical` unchanged.
```

Honestly not sure what is going on here because I am not manually providing the `getContentAnchorEl` props. Must have something to do with how `Menu` uses the `Popover` component internally. I just followed what it said and nothing breaks.

AbleLincoln: See https://github.com/mui-org/material-ui/blob/3a64fe16ff1bd3a72bf0d49d74eba280a839c2eb/packages/material-ui/src/Menu/Menu.js#L62

tasn: Cool, I believe you. All good. :)

tasn:
Btw, I changed the commit message a bit, we use the imperative form in the commit messages. So "improves" should be "improve".

AbleLincoln:
Ah I'll keep that in mind for future commits, thank you

PR #106, opened 2020-03-23
Tasks: add search functionality

AbleLincoln:
This PR also moves the quick actions/menu (search, filter, sort) to their own toolbar above quick add. Things were getting a bit crammed all on one line.

Also, I know we discussed earlier about having search be it's on separate page. The solution I went with here just uses the regular task list and basically the search term as another filter. The reason I chose this was for ease of integration. But if this still doesn't seem like the optimal solution we can discuss.


src/components/Tasks/QuickAdd.tsx
```diff
@@ -51,6 +58,7 @@ function QuickAdd(props: PropsType) {
         fullWidth
         value={title}
         onChange={handleChange}
+        InputLabelProps={{ className: useStyles().input }}
```
tasn: Please move `useStyles()` to the top of the function. All hooks have to be called in every render and in the same order. Moving this to year increases the chances a feature change will mess this up.

src/components/Tasks/TaskList.tsx
```diff
@@ -134,20 +157,29 @@ export default function TaskList(props: PropsType) {
...
   return (
     <Grid container spacing={4}>
-
-      <Grid item xs={3} style={{ borderRight: `1px solid ${theme.palette.divider}` }}>
-        <Sidebar tasks={potentialEntries} />
-      </Grid>
-
-      <Grid item xs>
+      <Grid item xs={12} style={{
+        backgroundColor: theme.palette.primary[500], padding: 0, paddingLeft: `calc(25% + ${theme.spacing(2)}px`,
```
tasn: What's up with the padding? Why such a complex calculation?

src/components/Tasks/Toolbar.tsx
```diff
@@ -20,23 +24,66 @@ import { StoreState } from '../../store';
 
 import Menu from '../../widgets/Menu';
 
+const transitionTimeout = 500;
```
tasn: I think the transition time is too long, should probably be 300 or maybe even less? I haven't tried modifying it, I just felt it was very slow.

src/components/Tasks/Toolbar.tsx
```diff
@@ -20,23 +24,66 @@ import { StoreState } from '../../store';
...
+
+const transitionStyles = {
+  entering: { visibility: 'visible', width: '100%', overflow: 'hidden' },
+  entered: { visibility: 'visible', width: '100%' },
+  exiting: { visibility: 'visible', width: '0%', overflow: 'hidden' },
+  exited: { visibility: 'hidden', width: '0%' },
+};
+
+const useStyles = makeStyles((theme) => ({
```
tasn: All of these styles feel like a hack to me. Why do you need such complex styling? Doesn't the transition component already cover the transition stuff? Is autocomplete still there? I didn't see.

src/components/Tasks/Toolbar.tsx
```diff
@@ -20,23 +24,66 @@ import { StoreState } from '../../store';
...
+  button: {
+    marginRight: theme.spacing(1),
+  },
+  autocomplete: {
+    transition: `width ${transitionTimeout}ms`,
+    marginRight: theme.spacing(1),
+  },
+  input: {
+    borderRadius: theme.shape.borderRadius,
```
tasn: All of this should probably be replaced by either the standard variant, or maybe the filled of the input box. Please don't make new styles, and instead use the styles provided by the MUI library.

src/components/Tasks/Toolbar.tsx
```diff
@@ -20,23 +24,66 @@ import { StoreState } from '../../store';
...
+    '&::before, &:hover::before, &::after': {
+      borderBottom: 0,
+    },
+
+    '& input': {
+      padding: '0.75em',
+    },
+  },
+  adornment: {
+    marginTop: '0!important',
```
tasn: `!important` is a code-smell. What do you need it for? Why do you need extra styling?

src/components/Tasks/Toolbar.tsx
```diff
@@ -49,11 +96,40 @@ export default function Toolbar(props: PropsType) {
...
+                  <SearchIcon />
+                </InputAdornment>
+              ),
+            }}
+          />
+        )}
+      </Transition>
+
+      <div className={classes.button}>
+        <IconButton size="small" onClick={toggleSearchField}>
```
tasn: This icon, like all the rest, are missing the `title` prop. Look at the refresh icon at the top right how it says "Refresh" if you hover with the mouse. This is not the case in your toolbar icons.

src/components/Tasks/QuickAdd.tsx
```diff
@@ -10,9 +10,16 @@ import ICAL from 'ical.js';
 import uuid from 'uuid';
 
 import TextField from '@material-ui/core/TextField';
+import { makeStyles } from '@material-ui/core/styles';
 
 import { TaskType, PimType, TaskStatusType } from '../../pim-types';
 
+const useStyles = makeStyles({
```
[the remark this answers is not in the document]
AbleLincoln: Sorry, actually I was playing around with different UI before I landed on this one. I think I forgot to remove some styles. This isn't needed anymore

src/components/Tasks/TaskList.tsx
AbleLincoln: It's basically to align the search bar with the tasks lists. I actually think I just realized a much better, clearer way to do this. Let me change in next commit

src/components/Tasks/Toolbar.tsx
AbleLincoln: I am using filled, but I'm trying to make it look better with the toolbar. I basically have all these complex styles in order to make the design fit more with our UI. The MUI TextFields just don't look good outside of the context of like a form. And the API is pretty limited in how it allows you to styles these things.

AbleLincoln: To give you an idea of what I'm going for. Here is what I have now:
Screen Shot 2020-03-25 at 2 01 44 AM

Here are how the MUI variants look:

**Standard**
Screen Shot 2020-03-25 at 2 02 50 AM

**Filled**
Screen Shot 2020-03-25 at 2 03 09 AM

**Outlined**
Screen Shot 2020-03-25 at 2 03 24 AM

To me, they all kind of don't quite look right. I guess the "outlined" variant is closest to what I want. I guess I could try starting with that and modifying the styles as minimally as possible to get something looking good.

tasn: I think all of them are more or less the same. As in, none of them is 2x the other (than than filled that's really ugly).
I'd just stick with the standard. The thing is, it's just not worth it to do things in a non-standard way because they will break the next time we update MUI. I've been burnt before, it's just not worth it... :|

AbleLincoln: Ok, I understand. I just wish the library gave you more options for modifications. It's a great library that can do a lot of cool things, until you want to deviate slightly from standard stuff. Then it's way too complicated.

tasn: I think that's on purpose. They want people to follow MUI, not deviate from it.

src/components/Tasks/Toolbar.tsx
AbleLincoln: No longer necessary, now that I've cleaned up those styles.

src/components/Tasks/Toolbar.tsx
AbleLincoln: I guess that makes sense. Maybe I just don't like Material haha. Or at least I think as a design philosophy is too rigid. I understand it's good to have rules, but I also like when apps have their own personality.

Either way, it makes development soooo much easier and faster, which I appreciate.

tasn: I agree about the personality, but as you said, there's also a plus-side. Consider MUI as a fast iteration framework for now. Maybe in the future we can have a cohesive design for the app that deviates from MUI. :)

src/components/Tasks/Toolbar.tsx
```diff
@@ -20,23 +24,51 @@ import { StoreState } from '../../store';
...
+  exiting: { visibility: 'visible', width: '0%', overflow: 'hidden' },
+  exited: { visibility: 'hidden', width: '0%' },
+};
+
+const useStyles = makeStyles((theme) => ({
+  button: {
+    marginRight: theme.spacing(1),
+  },
+  textField: {
+    transition: `width ${transitionTimeout}ms`,
```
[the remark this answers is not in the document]
AbleLincoln: Not quite. From the `Transition` [documentation](http://reactcommunity.org/react-transition-group/transition): 
> By default the `Transition` component does not alter the behavior of the component it renders, it only tracks "enter" and "exit" states for the components. It's up to you to give meaning and effect to those states.

The component doesn't really do anything but track states. Transition styles have to be manually applied.

src/components/Tasks/TaskList.tsx
```diff
@@ -96,13 +105,32 @@ interface PropsType {
...
   const settings = useSelector((state: StoreState) => state.settings.taskSettings);
   const { filterBy, sortBy } = settings;
   const theme = useTheme();
+  const classes = useStyles();
 
   const potentialEntries = React.useMemo(
-    () => props.entries.filter((x) => showCompleted || !x.finished),
-    [showCompleted, props.entries]
+    () => {
+      const entries = props.entries.filter((x) => showCompleted || !x.finished);
```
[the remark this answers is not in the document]
AbleLincoln: Oh right, forgot. Do you still want me to handle greying completed tasks out?

src/components/Tasks/Toolbar.tsx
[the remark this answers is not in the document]
tasn: So why do you need to pass the timeout to it?

src/components/Tasks/TaskList.tsx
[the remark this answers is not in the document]
tasn: Nah, that can be a future PR as it applies to every time the showCompleted is shown, not just in the search case. Not sure greying out is the right way too. Maybe how it is now (ticked checkbox) is sufficient even.

src/components/Tasks/Toolbar.tsx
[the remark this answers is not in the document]
AbleLincoln: To the stye or the component?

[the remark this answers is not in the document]
tasn: Why does the Transition component get a timeout if it doesn't actually do the transition? That's what I'm not getting.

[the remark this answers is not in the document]
AbleLincoln: Basically all the library is doing is helping apply different inline styles at different times. The actual animation of the transition effect comes from CSS. I believe the component needs to know the timeout so it flips states at the right time. The style needs it because that determines how long the actual transition animation lasts. Hypothetically these both could be different values, but that would just cause the transitions to look weird and/or cut off.

This is just how the library does it. I know it's a little clunky.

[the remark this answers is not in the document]
AbleLincoln: The `Transition` component allows you to apply different styles for 4 different phases of a transition: entering, entered, exiting, exited. It's not really doing any CSS transitions on it's own. Just transitioning between those states.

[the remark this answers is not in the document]
AbleLincoln: Haha I'm having a tough time explaining it, but the [documentation](http://reactcommunity.org/react-transition-group/transition) is pretty clear.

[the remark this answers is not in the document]
tasn: Cool, as long as it makes sense, I'm fine with that. Just found it odd...

AbleLincoln:
How's this?

src/components/Tasks/Toolbar.tsx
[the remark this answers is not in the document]
AbleLincoln: I know, they way the library works is a little strange.

PR #108, opened 2020-03-25
RRule: handle undefined options.byday

AbleLincoln:
RRule previously caused a crash on mount because `options.byday` is initially undefined.
`Select` component with prop `multiple` requires `value` to be an array.

### DIFF
--- a/src/widgets/RRule.tsx
+++ b/src/widgets/RRule.tsx
@@ -176,7 +176,7 @@ export default function RRule(props: PropsType) {
           <FormControl>
             <InputLabel>Weekdays</InputLabel>
             <Select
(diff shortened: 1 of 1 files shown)


src/widgets/RRule.tsx
```diff
@@ -176,7 +176,7 @@ export default function RRule(props: PropsType) {
           <FormControl>
             <InputLabel>Weekdays</InputLabel>
             <Select
-              value={sanitizeByDay(options.byday)}
+              value={sanitizeByDay(options.byday) ?? []}
```
tasn: Please move it into sanitizeByDay which is in charge of sanitizing exactly these kind of things.

PR #110, opened 2020-03-26
Tasks: make overdue secondary text red

AbleLincoln:


src/pim-types.ts
```diff
@@ -230,6 +230,10 @@ export class TaskType extends EventType {
     return this.dueDate && moment(this.dueDate.toJSDate()).isSameOrBefore(moment(), 'day');
   }
 
+  get overdue() {
+    return !!this.dueDate && moment(this.dueDate.toJSDate()).isBefore(moment(), 'day');
```
tasn: This function is wrong. :)

There are two cases here:
1. `dueDate` is a date (rather than datetime) - then this is correct.
2. 'dueDate` is a datetime - then this is wrong, and you should just compare the date as is...

src/widgets/List.tsx
```diff
@@ -46,6 +46,7 @@ interface ListItemPropsType {
   insetChildren?: boolean;
   nestedItems?: React.ReactNode[];
   selected?: boolean;
+  ListItemTextProps?: ListItemTextProps;
```
tasn: This is a very very leaky abstraction (not really an abstraction)...

Just call the prop `secondaryTextStyle` and pass just the style itself or something. Or if it's not a style, then just call it, `secondaryTextColor`.

src/pim-types.ts
AbleLincoln: ah yes I didn't think of that second case

AbleLincoln:
I believe that fixed everything. I also had to add some logic for displaying all day tasks due today as "Due today" (previously it would display time from 2am the day it was due).

src/components/Tasks/TaskListItem.tsx
```diff
@@ -57,7 +57,8 @@ export default React.memo(function TaskListItem(props: PropsType) {
   return (
     <ListItem
       primaryText={title}
-      secondaryText={task.dueDate && `Due ${moment().to(task.dueDate.toJSDate())}`}
+      secondaryText={task.dueDate && `Due ${task.dueDate.isDate && task.dueToday ? 'today' : moment(task.dueDate.toJSDate()).fromNow()}`}
```
tasn: 1. This change is unrelated to the PR, please don't shove in unrelated changes.
2. This is not desirable behaviour. The problem is that this doesn't update every time the clock ticks, so it gets stale very quickly... That's generally the problem with `fromNow`... To solve it, you would need to create a widget that automatically updates as the clock ticks (with configurable ticking resolution). It's not hard, just pointless for now...

src/pim-types.ts
```diff
@@ -230,6 +230,16 @@ export class TaskType extends EventType {
     return this.dueDate && moment(this.dueDate.toJSDate()).isSameOrBefore(moment(), 'day');
   }
 
+  get overdue() {
+    if (!this.dueDate) {
+      return false;
+    }
```
tasn: Maybe I'm just tired, but this piece of code looked so very confusing to me.
First of all, having the separate ifs here instead of `if ... else if ... else`.
Or even better, move the shared parts a bit above, so:
```
const dueDate = moment(this.dueDate.toJSDate());
const now = moment();
return (this.dueDate.isDate) ? dueDate.isBefore(now, 'day') : dueDate.isBefore(now);
```

tasn: Though, coming to think about it, I think that this condition is not really needed. If it has no time, the time is automatically 00:00 and then before, is the same as before(..., 'day).
So just change the whole thing to:

```
return (this.dueDate) ? moment(this.dueDate.toJSDate()).isBefore(moment()) : false;
```

src/components/Tasks/TaskListItem.tsx
AbleLincoln: So should we just not track due date at all in the UI?

src/pim-types.ts
AbleLincoln: But then this marks tasks overdue right when the day they are due starts. Like if a task is due on April 1, it will be be overdue on April 1. Is this how we want it to work? I would think that a due date without a time shouldn't be overdue until the day after.

src/components/Tasks/TaskListItem.tsx
tasn: No, due date is fine, and should be shown, the problem is only with "fromNow". The moment you show the actual date, it's fine.

src/pim-types.ts
tasn: Ah, and the "day" variant doesn't behave this way? Sure then, keep the differentiation then, though my first comment still applies.

src/components/Tasks/TaskListItem.tsx
AbleLincoln: Ok, I can change it to just show the due date. I just want to point out thought that the old code had already been showing the time to the due date. This `moment().to(task.dueDate.toJSDate())` is the same as this `moment(task.dueDate.toJSDate()).fromNow()`. 

I can get rid of that though, if you'd like.

tasn: Oh, that was a mistake most likely. Yeah, remove the `.to`, just show the time.

AbleLincoln: Ok, I will revert this change for this PR, and then fix that in a separate PR.

AbleLincoln:
Ok, should be much cleaner now.

PR #115, opened 2020-03-30
Tasks: add show hidden feature

AbleLincoln:


src/components/Tasks/TaskList.tsx
```diff
@@ -126,10 +127,12 @@ export default function TaskList(props: PropsType) {
         }).search(searchTerm);
         return result.map((x) => x.item);
       } else {
-        return props.entries.filter((x) => showCompleted || !x.finished);
+        const notCompleted = (x: TaskType) => showCompleted || !x.finished;
+        const notHidden = (x: TaskType) => showHidden || !x.hidden;
+        return props.entries.filter((x) => notCompleted(x) && notHidden(x));
```
tasn: Regarding the `notCompleted` and `notHidden`: as I told you before, we prefer named functions over anonymous ones.
Though in this case in particular, I think just put both inside the `.filter` directly rather than having separate functions. I'm sure/hope V8 is smart enough to inline these functions anyway, but since separating them doesn't add any clarity anyway, may as well just inline them yourself.

AbleLincoln: Like this?
```js
return props.entries.filter((x) => (showCompleted || !x.finished) && (showHidden || !x.hidden));
```

tasn: Yup, exactly.

tasn:
Great, thanks! :)
Btw, feel free to throw multiple PRs at me in parallel, you don't need to wait for my feedback with one before submitting the other (assuming you have the code ready). Though anyway, I often reply quickly so I guess it doesn't really matter...

AbleLincoln:
Haha alright I'll just spam you with PRs. Although sometimes I am waiting on specific code to be merged in so I can use it in another branch. Do you have a solution for how I can work around this?

tasn:
Yeah, for those I guess it's easiest to just wait and do them in order. Especially since more often than not, I have a quick turnaround.

tasn:
Sorry I missed it before, I now realised that the the "show hidden" toggle is not aligned to the right. Could you please fix that?

PR #116, opened 2020-03-30
UI: pretty error messages

AbleLincoln:
Add Toast component for displaying error messages in a more prominent way

### DIFF
--- a/src/components/EventEdit.tsx
+++ b/src/components/EventEdit.tsx
@@ -139,6 +139,7 @@ class EventEdit extends React.PureComponent<PropsType> {
     this.onDeleteRequest = this.onDeleteRequest.bind(this);
     this.toggleRecurring = this.toggleRecurring.bind(this);
     this.handleRRuleChange = this.handleRRuleChange.bind(this);
+    this.handleCloseToast = this.handleCloseToast.bind(this);
   }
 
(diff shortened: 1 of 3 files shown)


AbleLincoln:
Dang I always forget the license. I wonder if we can make a pre-commit hook for it...

tasn:
The problem with pre-commit hooks is that they are essentially remote code executions exploits. That is, if I can still a pre-commit hooks for you (as in automatically), I can run whatever code I want on your system.
You can, if you want add one for yourself, and maybe we can put them in the repo. I wouldn't object to it, but I wouldn't bother doing it myself either. :)

tasn:
I also noticed earlier today that the error messages of the login screen are ugly as hell. This looks like a nice addition, maybe I should use that. :)

AbleLincoln:
> The problem with pre-commit hooks is that they are essentially remote code executions exploits. That is, if I can still a pre-commit hooks for you (as in automatically), I can run whatever code I want on your system.

Oh wild I didn't even think of that. That's actually kind of funny how dangerous it is. I guess the hook could really be any code, huh?

I could try taking a stab at it. I've never written a git hook before but I'm sure it's not too difficult.

tasn:
Yeah, any script or executable! Normally you would put such a check on the server (and what I did with other codebases in the past), but I don't think github support such hooks server side.

PR #120, opened 2020-04-04
pim-types: change clone to deep clone using to/fromString

### LINKED ISSUE #119: Bug: Event/Task `clone` doesn't deep clone
The [`clone` function](https://github.com/etesync/etesync-web/blob/77f734a227cb2b08c6b3c245753e73bd8b11c309/src/pim-types.ts#L125-L130) uses ical's `toJson` function to make a copy. Unfortunately, this function just returns the object representation of a component, it doesn't actually make a json copy as one would expect. This is dangerous: as the documentation says, ["The returned object is a live jCal object and should be cloned if modified"](http://mozilla-comm.github.io/ical.js/api/ICAL.Component.html#toJSON).

This is causing problems for instance in the Tasks recurrence code. When I handle completing a recurring task, I need to use `copy` to create the next occurrence, modify it, and save it. Since the underlying jCal objects for both tasks are really just the same object, anything I modify on the "clone" is modified on the original (only in memory, until I inevitably save them).

AbleLincoln:
fixes #119

### DIFF
--- a/src/pim-types.ts
+++ b/src/pim-types.ts
@@ -123,7 +123,7 @@ export class EventType extends ICAL.Event implements PimType {
   }
 
   public clone() {
-    const ret = new EventType(new ICAL.Component(this.component.toJSON()));
+    const ret = new EventType(ICAL.Component.fromString(this.component.toString()));
     ret.color = this.color;
     return ret;
   }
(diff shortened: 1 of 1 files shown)


tasn:
This doesn't feel right... :|
Is there an easy way to just deep clone? If not, no worries, we can use this, but :nauseated_face: :)

AbleLincoln:
I know it's weird, but the library doesn't provide any better way to do a true deep clone. We could write some kind of deep cloning function, but that seems like it would be pretty brittle.

AbleLincoln:
There are stringify/parse functions, but this is pretty much the same thing 🤷🏼‍♀️

tasn:
:broken_heart:

PR #121, opened 2020-04-04
Tasks: fix priority sorting

AbleLincoln:
Priority sorting had some inconsistencies. 
The Infinity was causing tasks to not be sorted (Infinity - Infinity = NaN).
Adding sortByDueDate also ensures better sorting when tasks have the same priority


tasn:
Please, next time, split these two unrelated changes to separate commits. I'll go ahead and merge this one though...

Issue #117, opened 2020-03-31
Feature request: clear completed tasks


AbleLincoln:
I would like to add a way delete all completed tasks.

Does the library currently provide any way to perform a batch delete?

tasn:
Yeah, it does, though not as part of the `PIM` abstraction, you will have to implement it yourself.
Take a look at `etesync-helpers.ts` you have there the current wrapper we use (`addJournalEntry`), you should create another one (or modify this one?) to accept an array instead of just one and then create multiple entries and call `addEntries` with the array.

Things to keep in mind:
1. `prevUid` is the uid of the previous item. So when you create multiple items, you need to update it every time, and every time set it to the uid of the entry you are just going to send.
2. You probably don't want to upload 400 changes at once, so you should loop over it in chunks of e.g. 30 and push them each. You can take a look at `helpers.tsx` in the ios repo for a helper function to split an array to a chunks iterator. You can also look at `sync/SyncManagerBase.ts` there for how we push in chunks.

Let me know if you need anything.

PR #122, opened 2020-04-05
Tasks: support reccuring tasks

### LINKED ISSUE #107: Add Support for Recurring Tasks
It would be great if etesync could support recurring tasks along with recurring calendar events. Its a part of the spec but so few clients implement it. And thats a shame because its such a useful feature.  If you guys implement this, i guarantee you it would put you on the map

AbleLincoln:
Resolves #107

### DIFF
--- a/src/components/Tasks/TaskList.tsx
+++ b/src/components/Tasks/TaskList.tsx
@@ -6,14 +6,14 @@ import * as React from 'react';
 import * as EteSync from 'etesync';
 
 import { List } from '../../widgets/List';
-import Toast from '../../widgets/Toast';
+import Toast, { PropsType as ToastProps } from '../../widgets/Toast';
 
 import { TaskType, PimType, TaskStatusType } from '../../pim-types';
 import Divider from '@material-ui/core/Divider';
 import Grid from '@material-ui/core/Grid';
 import { useTheme, makeStyles } from '@material-ui/core/styles';
 
-import { useSelector } from 'react-redux';
+import { useSelector, useDispatch } from 'react-redux';
 
 import Fuse from 'fuse.js';
 
@@ -22,8 +22,12 @@ import Sidebar from './Sidebar';
 import Toolbar from './Toolbar';
 import QuickAdd from './QuickAdd';
 
-import { StoreState } from '../../store';
+import { StoreState, UserInfoData, CredentialsData } from '../../store';
 import { formatDate } from '../../helpers';
+import { SyncInfo } from '../../SyncGate';
+import { fetchEntries } from '../../store/actions';
+import { Action } from 'redux-actions';
+import { addJournalEntries } from '../../etesync-helpers';
 
 function sortCompleted(a: TaskType, b: TaskType) {
   return (!!a.finished === !!b.finished) ? 0 : (a.finished) ? 1 : -1;
@@ -104,30 +108,71 @@ interface PropsType {
   collections: EteSync.CollectionInfo[];
   onItemClick: (entry: TaskType) => void;
   onItemSave: (item: PimType, journalUid: string, originalItem?: PimType) => Promise<void>;
+  syncInfo: SyncInfo;
+  userInfo: UserInfoData;
+  etesync: CredentialsData;
 }
 
 export default function TaskList(props: PropsType) {
   const [showCompleted, setShowCompleted] = React.useState(false);
   const [showHidden, setShowHidden] = React.useState(false);
   const [searchTerm, setSearchTerm] = React.useState('');
-  const [info, setInfo] = React.useState('');
+  const [toast, setToast] = React.useState<{ message: string, severity: ToastProps['severity'] }>({ message: '', severity: undefined });
   const settings = useSelector((state: StoreState) => state.settings.taskSettings);
   const { filterBy, sortBy } = settings;
   const theme = useTheme();
   const classes = useStyles();
+  const dispatch = useDispatch();
 
-  const handleToggleComplete = async (task: TaskType, completed: boolean) => {
+  const handleToggleComplete = (task: TaskType, completed: boolean) => {
     const clonedTask = task.clone();
     clonedTask.status = completed ? TaskStatusType.Completed : TaskStatusType.NeedsAction;
 
     const nextTask = completed ? task.getNextOccurence() : null;
 
-    await props.onItemSave(clonedTask, (task as any).journalUid, task);
+    const syncJournal = props.syncInfo.get((task as any).journalUid);
 
-    if (nextTask) {
-      await props.onItemSave(nextTask, (task as any).journalUid);
-      setInfo(`${nextTask.title} rescheduled for ${formatDate(nextTask.startDate ?? nextTask.dueDate)}`);
+    if (syncJournal === undefined) {
+      setToast({ message: 'Could not sync.', severity: 'error' });
+      return;
     }
+
+    const journal = syncJournal.journal;
+
+    let prevUid: string | null = null;
+    let last = syncJournal.journalEntries.last() as EteSync.Entry;
+    if (last) {
+      prevUid = last.uid;
+    }
+
+    dispatch<any>(fetchEntries(props.etesync, journal.uid, prevUid))
+      .then((entriesAction: Action<EteSync.Entry[]>) => {
+        last = entriesAction.payload!.slice(-1).pop() as EteSync.Entry;
+
+        if (last) {
+          prevUid = last.uid;
+        }
+
+        const changeTask = [EteSync.SyncEntryAction.Change, clonedTask.toIcal()];
+
+        const updates = [];
+        updates.push(changeTask as [EteSync.SyncEntryAction, string]);
+
+        if (nextTask) {
+          const addNextTask = [EteSync.SyncEntryAction.Add, nextTask.toIcal()];
+          updates.push(addNextTask as [EteSync.SyncEntryAction, string]);
+        }
+
+        return dispatch(addJournalEntries(props.etesync, props.userInfo, journal, prevUid, updates));
+      })
+      .then(() => {
+        if (nextTask) {
+          setToast({ message: `${nextTask.title} rescheduled for ${formatDate(nextTask.startDate ?? nextTask.dueDate)}`, severity: 'success' });
+        }
+      })
+      .catch(() => {
+        setToast({ message: 'Failed to save changes. This may be due to a network error.', severity: 'error' });
+      });
   };
 
   const potentialEntries = React.useMemo(
@@ -212,8 +257,8 @@ export default function TaskList(props: PropsType) {
         </List>
       </Grid>
 
-      <Toast open={!!info} severity="info" onClose={() => setInfo('')} autoHideDuration={3000}>
-        {info}
+      <Toast open={!!toast.message} severity={toast.severity} onClose={() => setToast({ message: '', severity: undefined })} autoHideDuration={3000}>
+        {toast.message}
       </Toast>
     </Grid>
   );
--- a/src/components/Tasks/TaskList.tsx
+++ b/src/components/Tasks/TaskList.tsx
@@ -6,8 +6,9 @@ import * as React from 'react';
 import * as EteSync from 'etesync';
 
 import { List } from '../../widgets/List';
+import Toast from '../../widgets/Toast';
 
-import { TaskType, PimType } from '../../pim-types';
+import { TaskType, PimType, TaskStatusType } from '../../pim-types';
 import Divider from '@material-ui/core/Divider';
 import Grid from '@material-ui/core/Grid';
 import { useTheme, makeStyles } from '@material-ui/core/styles';
@@ -22,6 +23,7 @@ import Toolbar from './Toolbar';
 import QuickAdd from './QuickAdd';
 
 import { StoreState } from '../../store';
+import { formatDate } from '../../helpers';
 
 function sortCompleted(a: TaskType, b: TaskType) {
   return (!!a.finished === !!b.finished) ? 0 : (a.finished) ? 1 : -1;
@@ -108,10 +110,25 @@ export default function TaskList(props: PropsType) {
   const [showCompleted, setShowCompleted] = React.useState(false);
   const [showHidden, setShowHidden] = React.useState(false);
   const [searchTerm, setSearchTerm] = React.useState('');
+  const [info, setInfo] = React.useState('');
   const settings = useSelector((state: StoreState) => state.settings.taskSettings);
   const { filterBy, sortBy } = settings;
   const theme = useTheme();
   const classes = useStyles();
+
+  const handleToggleComplete = async (task: TaskType, completed: boolean) => {
+    const clonedTask = task.clone();
+    clonedTask.status = completed ? TaskStatusType.Completed : TaskStatusType.NeedsAction;
+
+    const nextTask = completed ? task.getNextOccurence() : null;
+
+    await props.onItemSave(clonedTask, (task as any).journalUid, task);
+
+    if (nextTask) {
+      await props.onItemSave(nextTask, (task as any).journalUid);
+      setInfo(`${nextTask.title} rescheduled for ${formatDate(nextTask.startDate ?? nextTask.dueDate)}`);
+    }
+  };
 
   const potentialEntries = React.useMemo(
     () => {
@@ -156,7 +173,7 @@ export default function TaskList(props: PropsType) {
         key={uid}
         entry={entry}
         onClick={props.onItemClick}
-        onSave={props.onItemSave}
+        onToggleComplete={(completed: boolean) => handleToggleComplete(entry, completed)}
       />
     );
   });
@@ -194,6 +211,10 @@ export default function TaskList(props: PropsType) {
           {itemList}
         </List>
       </Grid>
+
+      <Toast open={!!info} severity="info" onClose={() => setInfo('')} autoHideDuration={3000}>
+        {info}
+      </Toast>
     </Grid>
   );
 }
--- a/src/components/Tasks/TaskListItem.tsx
+++ b/src/components/Tasks/TaskListItem.tsx
@@ -3,7 +3,7 @@
 
 import * as React from 'react';
 
-import { TaskType, TaskStatusType, PimType, TaskPriorityType } from '../../pim-types';
+import { TaskType, TaskPriorityType } from '../../pim-types';
 import { ListItem } from '../../widgets/List';
 
 import Checkbox from '@material-ui/core/Checkbox';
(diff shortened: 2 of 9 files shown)


AbleLincoln:
Here is the code for the recurrence feature. It was by far the most complicated thing I've had to do yet. All the features are here and working, but I think the code could be improved a bit. I wanted to make the PR so we can discuss.

tasn:
I'm so sorry, I thought I already reviewed it! I'm going to do it now.

src/types/ical.js.d.ts
```diff
@@ -176,10 +176,19 @@ declare module 'ical.js' {
     public bysetpos?: number[];
   }
 
+  export class RecurIterator {
```
tasn, line 179: For all of this file, please copy over the file how it is from the ios repo. This is already mostly (fully?) implemented there, so no need to implement it here differently.

src/components/Tasks/TaskListItem.tsx
```diff
@@ -36,26 +36,40 @@ interface PropsType {
...
   onSave: (item: PimType, journalUid: string, originalItem?: PimType) => Promise<void>;
+  onUpdate: (message: string) => void;
 }
 
 export default React.memo(function TaskListItem(props: PropsType) {
   const {
     entry: task,
     onClick,
     onSave: save,
+    onUpdate: update,
```
tasn, line 47: Why rename this function?

src/components/Tasks/TaskListItem.tsx
```diff
@@ -36,26 +36,40 @@ interface PropsType {
   entry: TaskType;
   onClick: (task: TaskType) => void;
   onSave: (item: PimType, journalUid: string, originalItem?: PimType) => Promise<void>;
+  onUpdate: (message: string) => void;
```
tasn, line 39: I really don't think this is the way to do it... First of all, `onUpdate` sounds like a generic name which can mislead you into thinking this is actually just a generic action, when in fact, the message parameter is very tied to what you are trying to do.
Also, the fact that you also only call this when rescheduling a task, and not really on an update, is quite misleading.
I don't think this is the right way to do it. You maybe just want a redux queue for toasts and a toast presenter somewhere high up, that just processeses the queue and presents toasts.

src/components/Tasks/TaskListItem.tsx
```diff
@@ -36,26 +36,40 @@ interface PropsType {
...
+
+    if (nextTask) {
+      await save(nextTask, (task as any).journalUid);
+      update(`${nextTask.title} rescheduled for ${formatDate(nextTask.startDate ?? nextTask.dueDate)}`);
+    }
   }
 
+  const dueDateText = task.dueDate ? `Due ${formatDate(task.dueDate)}` : '';
+  const freqText = task.rrule ? `(repeats ${task.rrule.freq.toLowerCase()})` : '';
+  const secondaryText = `${dueDateText} ${freqText}`;
```
tasn, line 67: This looks odd if there's no due date...

src/pim-types.ts
```diff
@@ -255,6 +264,57 @@ export class TaskType extends EventType {
...
+    nextTask.uid = uuid.v4();
+    if (nextStartDate) {
+      nextTask.startDate = nextStartDate;
+    }
+    if (nextDueDate) {
+      nextTask.dueDate = nextDueDate;
+    }
+
+    if (this.rrule.isByCount()) {
+      rrule.count = (this.rrule.count as number) - 1;
```
tasn, line 309: Why is this cast needed?

src/pim-types.ts
```diff
@@ -255,6 +264,57 @@ export class TaskType extends EventType {
...
+      if (next.compare(ICAL.Time.now()) > 0) {
+        break;
+      }
+    }
+
+    if (!next) {
+      return null; // end of reccurence
+    }
+
+    const nextStartDate = this.startDate && next;
```
tasn, line 292: I don't get this `&& next`... You already test that next is not null a few lines above, and why have this in the first place?

src/pim-types.ts
```diff
@@ -255,6 +264,57 @@ export class TaskType extends EventType {
...
+    const nextDueDate = this.dueDate && next;
+    if (nextStartDate && nextDueDate) {
+      const offset = (this.dueDate as ICAL.Time).subtractDateTz(this.startDate);
+      nextDueDate.addDuration(offset);
```
tasn, line 296: Maybe I'm missing something, but isn't nextDueDate a boolean? How is this even going forward?

src/pim-types.ts
```diff
@@ -255,6 +264,57 @@ export class TaskType extends EventType {
...
+    }
+
+    if (!next) {
+      return null; // end of reccurence
+    }
+
+    const nextStartDate = this.startDate && next;
+    const nextDueDate = this.dueDate && next;
+    if (nextStartDate && nextDueDate) {
+      const offset = (this.dueDate as ICAL.Time).subtractDateTz(this.startDate);
```
tasn, line 295: Why this cast?

src/types/ical.js.d.ts
AbleLincoln: should I copy the entire file or just the relevant parts?

src/components/Tasks/TaskListItem.tsx
AbleLincoln: I also felt that this was clunky but wasn't sure how to handle it. I like the idea of a redux queue for processing toasts. Maybe this is something that should happen in a separate PR then?

src/components/Tasks/TaskListItem.tsx
AbleLincoln: Really? I actually like how it's displayed, I think it gives context for the task. Should I remove the parenthesis if there is no due date? Or just remove it entirely if there is no due date?

src/types/ical.js.d.ts
AbleLincoln: Just looked at the code and it doesn't include the same typings that I added. Actually, I think this file was initially taken from the iOS repo because it is exactly the same (except for what I added). I had to add extra types that were needed for this code.

src/pim-types.ts
AbleLincoln: `rrule.count` can be `null`. `rrule.isByCount` actually checks if this number is `null`, but Typescript doesn't know that so it complains. I guess I could just change the check to use `rrule.count` instead of `rrule.isByCount`.

src/pim-types.ts
AbleLincoln: I'm using short-circuit evaluation. `next` is a Time. What I am trying to do is assign `nextStartDate` to `next` only if the `startDate` exists. Same with `endDate`. This might not be clear from my code though

src/pim-types.ts
AbleLincoln: No, it's either a `Time` (if start/dueDate exist) or it's undefined. I only need to calculate the offset if both are defined.

src/pim-types.ts
AbleLincoln: It could be a ternary instead: `const nextStartDate = this.startDate ? next : undefined`

src/pim-types.ts
AbleLincoln: `this.dueDate` can be `undefined`. This branch will only run if `nextDueDate` is defined (which means `this.dueDate` is defined), but Typescript doesn't know that so it complains.

src/components/Tasks/TaskListItem.tsx
AbleLincoln: Another idea I was thinking of is should the `TaskList` component just handle all the saving/updating of the items? That way any notification displaying can be contained within that component, and it also takes a lot of the functionality out of the `TaskListItems` and makes them purely presentational.

src/types/ical.js.d.ts
tasn: Sure, you can add extra types, but first copy it over, because with your changes the file diverged, and you implemented some things differently.

src/components/Tasks/TaskListItem.tsx
tasn: Yeah, this idea sounds good too. I don't like the PIM abstraction I did anyway, it was stupid and was more trouble than it's worth. Feel free to change that.

src/components/Tasks/TaskListItem.tsx
tasn: Sorry, let me clarify, I just realise that my message was really bad.
When a task has no dueDate set, there's a redundant space before the `freqText` which looks odd.

src/pim-types.ts
tasn: the correct way to count optional to non-optional in typescript is the `!` operator. So `this.rrule.count! - 1`.

Alternatively, why not just check if `this.rrule.count` is not null instead of using `isByCount`? I'm sure there's a reason, as they created this function, I'm just wondering.

src/pim-types.ts
tasn: Oh I forgot typescript does that. I think this code is not very obvious. Maybe your suggested alternative is better, yeah.

src/pim-types.ts
tasn: Yeah, my bad.

src/pim-types.ts
```diff
@@ -255,6 +264,57 @@ export class TaskType extends EventType {
...
+
+    const rrule = this.rrule.clone();
+
+    if (rrule.count && rrule.count <= 1) {
+      return null; // end of reccurence
+    }
+
+    rrule.count = null; // clear count so we can iterate as many times as needed
+    const recur = rrule.iterator(this.startDate ?? this.dueDate);
+    let next = recur.next();
```
tasn, line 281: Also, maybe consider renaming this variable, because it's a bit non-obvious. I'm sorry for the nit-picking here, it's just something I keep on coming back to and getting confused by when reviewing this.

src/pim-types.ts
```diff
@@ -255,6 +264,57 @@ export class TaskType extends EventType {
...
   }
+
+  public getNextOccurence(): TaskType | null {
+    if (!this.isRecurring()) {
+      return null;
+    }
+
+    const rrule = this.rrule.clone();
+
+    if (rrule.count && rrule.count <= 1) {
```
tasn, line 275: Also, I see here you are using the `rrule.count != null` rather than `isByCount`... See my other comment.

src/pim-types.ts
tasn: As I said elsewhere, the correct way of doing this in typescript is the `!` operator.
Also, check to see if typescript is smart enough to understand it is indeed not null once you change to the trenary operator as you suggested. If it is, great, if not, do the above.

src/types/ical.js.d.ts
AbleLincoln: Sorry, but I am a little confused. As far as I can tell, the files were exactly the same before I added anything. And the things I have implemented here don't exist in the iOS file, so I'm not quite sure what I should be copying over?

The things I've implemented differently are purposefully different. I'm not sure if you are referring to the divergence of this `RecurIterator` vs the `RecurExpansion`, but these are intentionally different. I'm going off of the icaljs docs here. [`Event.iterator`](https://mozilla-comm.github.io/ical.js/api/ICAL.Event.html#iterator) and [`Recur.iterator`](https://mozilla-comm.github.io/ical.js/api/ICAL.Recur.html#iterator) return different objects. The former returns a `RecurExpansion` and the latter returns a `RecurIterator`. I am using the latter in this PR.

src/components/Tasks/TaskListItem.tsx
AbleLincoln: Ah ok I can make sure that's removed if not necessary

src/pim-types.ts
AbleLincoln: > Alternatively, why not just check if this.rrule.count is not null instead of using isByCount? I'm sure there's a reason, as they created this function, I'm just wondering.

Yeah I was just trying to be semantic with my use of the library. Actually, looking at the source code the `isByCount` function does more than I need it to. It `count` and `until`, which I don't want. So I'll just use `this.rrule.count`.

src/pim-types.ts
AbleLincoln: Renaming `next`? Would `nextRecurrence` be better?

src/pim-types.ts
tasn: Great.

src/pim-types.ts
tasn: Yeah, that sounds better.

src/types/ical.js.d.ts
tasn: My bad, sorry. I didn't notice this was on Recur instead of Event.

src/components/Tasks/TaskListItem.tsx
AbleLincoln: haha I don't mind the abstraction, but it has caused some issues. I don't think I'll go so far as reworking that in this PR though, but maybe it is something to think about for the future.

src/components/Tasks/TaskListItem.tsx
AbleLincoln: I think I just wanted to keep it consistent, but I've removed these functions now

src/components/Tasks/TaskListItem.tsx
AbleLincoln: Looking at this in my browser (Firefox) and the leading white-space is removed when the component is rendered. Where are you seeing this issue?

AbleLincoln:
Hey, I made a few changes re: your comments. I cleaned up those sections of `pim-types` that were unclear. I've also hoisted the saving/updating logic from the `TaskListItem` up to its parent `TaskList`. This makes the `TaskListItem` muuuuuuch cleaner (I was having some issues with the component rerendering in the middle its save function, which is no longer an issue).

AbleLincoln:
The biggest issue I'm having now is that the UI is just kind of slow and janky. I'm really not sure on the best way to handle this. I've thought of including some animations so it's more clear what's going on. Not sure... I would love to hear your suggestions 🙂

src/components/Tasks/TaskListItem.tsx
tasn: I can't see it either. Never mind then. :|
Sorry for the noise.

tasn:
Why is it slow and janky? Do you have any information? Which part? Is it because of the network roundtrip or what? Can't comment without knowing where it's going on.

tasn:
I haven't tested it, but if it works, great, looks good to me. If you can just squash together related commits, that would be great. Then I'll be able to just merge it from the web UI.
Additionally, please let me know about the jankiness you mentioned, so I can test it locally.

AbleLincoln:
Yeah, sorry, I should have added a bit more context. There is just quite a lot of delay between clicking the checkbox and the entire action finishing, especially when it's a recurring task. When you check a recurring task it has to send an update for the original -> which causes a fetch of all tasks -> then you have to send a create for the new task -> which causes an entirely new fetch.

Throughout all this the UI isn't very clear. On checking the box, the app hangs for a second while it updates and fetches, then the task just disappears, then after another second the next task appears.

This is why I've added the info toast, just to clarify a bit what is going on. But I feel that the UI is not very responsive or fluid. I'm not sure what would be the best solution at this point. I'm not sure if we should batch the two update/fetch requests so that they both only happen once (kind of like we were talking about in #117 ).

tasn:
Ah yeah, that's something we need to take care of... In more than a few places the web ui depends on network latency (which is bad!).
As for what you described about two steps with a recurring task: this shouldn't happen! You should send both change entries together in one go, not two separate ones. Sending them separately is both much slower (as you discovered) and can lead to a bad state (the old task closed, and no new one created). This definitely needs to be addressed before this is merged. I'm sorry I missed this in my review. I thought this is why you wanted to remove the PIM abstraction and have the code directly here, so you can update both of them in one go...

AbleLincoln:
Sorry for the miscommunication, I wasn't even thinking about rewriting the code to upload in one go. But you're right that is definitely what should happen. I can refactor for this.

I'm still trying to understand fetching/syncing/adding works. In the PIM abstraction, I see that we do a [fetch](https://github.com/etesync/etesync-web/blob/84f7a11bbc3a8fd738a4038184f3ade109796cb3/src/Pim/index.tsx#L260) to get the most recent uid before sending an add. This makes sense, because we want to make sure the previous uid is most up to date. But how come a [few lines](https://github.com/etesync/etesync-web/blob/84f7a11bbc3a8fd738a4038184f3ade109796cb3/src/Pim/index.tsx#L256) above we _also_ use `syncJournal.journalEntries.last()` to get the previous uid?

tasn:
We fetch to get the most recent uid, as you said, nothing to add there. The reason why we get the previous uid before is because we only want to fetch from that uid onwards. That uid is the last one we *already have* locally, and we want it, so we can pass it to fetch so it will only fetch changes since that uid.

This code however, should remain the same, the only thing that should change is [the call to addJournalEntry](https://github.com/etesync/etesync-web/blob/84f7a11bbc3a8fd738a4038184f3ade109796cb3/src/Pim/index.tsx#L269) that should now become a call to a new function that adds the two entries (modify + add) at once as I described in a comment above.

AbleLincoln:
Ok this new code now sends the two updates at once using the method we discussed. I've added an `addJournalEntries` helper function that takes an array of tuples that are `(sync action, ical content)`. This has improved latency by a lot.

.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
+{
```
tasn, line 1: This file is here by mistake I assume.

src/components/Tasks/TaskList.tsx
```diff
@@ -104,30 +108,71 @@ interface PropsType {
...
+  userInfo: UserInfoData;
+  etesync: CredentialsData;
 }
 
 export default function TaskList(props: PropsType) {
   const [showCompleted, setShowCompleted] = React.useState(false);
   const [showHidden, setShowHidden] = React.useState(false);
   const [searchTerm, setSearchTerm] = React.useState('');
-  const [info, setInfo] = React.useState('');
+  const [toast, setToast] = React.useState<{ message: string, severity: ToastProps['severity'] }>({ message: '', severity: undefined });
```
tasn, line 120: How is this change related to this commit? I'll let it slide this time, but it feels like quite the unrelated change, which made this review harder than it should have been.

AbleLincoln: I felt it was related because I wanted to have error handling in the new sync code. It seemed weird to be using `setInfo` for an error, so I expanded it. I can split up the changes if you'd like, but the code is fairly coupled.

.vscode/launch.json
AbleLincoln: My bad. I keep meaning to add a gitignore rule for this.

src/components/Tasks/TaskList.tsx
tasn: No worries, it was just a bit confusing.

.vscode/launch.json
tasn: Feel free to open a PR for this.

src/components/Tasks/TaskList.tsx
tasn: P.S, I make some nit-picky comments every now and then, and I make the same comments to myself too. :P 
What I'm trying to say: don't worry about them too much. :)

AbleLincoln: Yeah I sometimes struggle with where to divide my commits, especially when developing things where two features are distinct yet rely on each other. It's like I don't want to push an incomplete feature, but I also don't want the changes to be too broad.

AbleLincoln:
Great, thank you! This one was a doozy, definitely the most difficult feature yet. Thank you for your patience.

tasn:
One comment: the error `'A recurring task must have either Hide Until or Due Date set!'` is showing in green instead of red... It should indicate failure, not success...

As for the previously mentioned UI lag: wow, it's very noticeable. It also seems to happen much after the network request. The network requests finishes and then it takes 1-2 more seconds. I suspect it may be something with the code, rather than related to network latency. This is not a blocker for this PR, as it seems unrelated, but it's definitely relevant.

AbleLincoln:
> One comment: the error 'A recurring task must have either Hide Until or Due Date set!' is showing in green instead of red... It should indicate failure, not success...

Whoops, I changed up the Toast API. I'll get this fixed

AbleLincoln:
> As for the previously mentioned UI lag: wow, it's very noticeable. It also seems to happen much after the network request. The network requests finishes and then it takes 1-2 more seconds. I suspect it may be something with the code, rather than related to network latency. This is not a blocker for this PR, as it seems unrelated, but it's definitely relevant.

Yeah, it's not great.... After this is merged in I'll do a performance audit with the aim to fix it for the following PR.

tasn:
Great job, thanks!

PR #130, opened 2020-05-31
Add an option for changing the default API URL during the build

saierd:
I am building a Docker image containing the EteSync server and the web app. With this change, I can simply set the default API URL to `/` so that the web client automatically connects to the location on which the Docker container is hosted. Otherwise the user would have to change the server address for every login.


tasn:
Merged, thanks!
Btw, if you are an IRC/Matrix user, please come and join us at https://www.etesync.com/community-chat/ I'm sure people would love to hear about what you are working on. :)

PR #139, opened 2020-07-04
Automatic refresh

### LINKED ISSUE #28: Implement automatic refresh
It's super easy, it's just a matter of adding a timer that calls fetchAll on an interval and making sure that it doesn't break anything when there's an error (e.g. when there's no internet connection).

ramzan:
Implement an auto refresh function that calls fetchAll every minute if the user is online and logged in.

Fixes #28


tasn:
Thank you so much, it's been a long awaited feature!

I'll mention it (and you) in the next blog update about the web client, I'm sure people will be thrilled to learn about it!

PR #143, opened 2020-07-15
Split name field

### LINKED ISSUE #80: ContactEdit: split to first/last/middle/etc names (like e.g. on Android) and correctly save 'n' field
The name field is currently not being updated when the name is changed. The reason for that is that we don't have the proper name fields in the UI (so those need to be added).
After we have those, we should correctly set it. Example how it's done is available in the iOS client, here: https://github.com/etesync/ios/blob/master/src/sync/helpers.ts#L572

@Tal-Leibman, could you maybe take a look?

ramzan:
Fixes #80

### DIFF
--- a/src/components/ContactEdit.tsx
+++ b/src/components/ContactEdit.tsx
@@ -130,6 +130,11 @@ class ContactEdit extends React.PureComponent<PropsType> {
   public state: {
     uid: string;
     fn: string;
+    namePrefix: string;
+    firstName: string;
+    middleName: string;
+    lastName: string;
+    nameSuffix: string;
     phone: ValueType[];
     email: ValueType[];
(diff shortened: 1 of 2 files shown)


tasn:
Looks great, and thanks a lot! I have just one comment: according to the standard the `n` field SHOULD be set, but doesn't have to be set when `fn` is set. So if we are no longer going to display `fn`, we need to deconstruct it into `n` when reading a contact, the same way you are reconstructing it when writing one. Does that makes sense?

Relevant part of the RFC: https://tools.ietf.org/html/rfc6350#section-6.2.2

ramzan:
OK, I think 06b0b5e should do it.

tasn:
Great, thanks a lot!

PR #147, opened 2020-07-17
Implement Event copying

### LINKED ISSUE #51: Duplicate/ Copy events in calendar
Will be useful to add copy/ duplicate feature for calendars. For regular appointments, like a doctor's, dentist, car service, etc., we can simply copy the event but change the date instead of having to re-enter everything each time.

ramzan:
Fixes #51

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -55,6 +55,7 @@ export const routeResolver = new RouteResolver({
       _id: {
         _base: ':itemUid',
         edit: 'edit',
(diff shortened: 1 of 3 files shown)


tasn:
Actually though, one though. Maybe instead of the copy icon and texting, could you have something that implies duplication? So the title would be "Duplicate" and the icon could either stay like copy (if relevant), or something better if there is one.

PR #148, opened 2020-07-17
Fix autoRefresh interrupting editing

### LINKED ISSUE #145: refresh resets
hello again.  i have come a cross an issue when using the web version: when editing a task, it refreshes every approx 10 seconds, thus resetting all text and changes (unless i save every few seconds).  

perhaps a solution would be to not refresh while viewing/editing a task?

ramzan:
The autoRefresh function now checks the end of the URL to prevent refreshing when the user is editing an item. 

Fixes #145

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -282,7 +282,11 @@ class App extends React.PureComponent {
   }
 
   private autoRefresh() {
-    if (navigator.onLine && this.props.credentials) { this.refresh() }
+    if (navigator.onLine && this.props.credentials &&
+      !(window.location.pathname.match(/.*\/(new|edit|copy)$/))) { 
+      this.refresh();
+    }
+
   }
 
   componentDidMount() {


tasn:
Thanks a lot!

I'm going to accept it now, as a hot fix, because it's already affecting user and is annoying, though I don't think it's the right way of solving it. This is very fragile and will break if we ever add another place with editing boxes.
There are two possible solutions I can think of:
1. Mark *inside* each view if it allows for auto-refresh or not - not great.
2. Change the way auto-refresh is done so it doesn't affect edit boxes - what I think is the right way.

The reason why this "block until synced" was implemented is because the initial sync was thrashing react with all the changes and was making everything slow, so one solution was to just remove all of those components until we fetched everything.

A better approach would be to remove all of the `fetchCount` checks everywhere, and when it comes to the sync info: only modify it when `fetchCount` is 0. So show the stale information while it's still syncing and only update the result when fetchCount is back to 0. Does this makes sense?

ramzan:
For some reason the issue seems to have disappeared for me. When I undo da22f79 and rebuild, the refresh occurs without interrupting the app. Does this happen for you?

tasn:
I was able to reproduce it before the change. I actually got bitten by it before I saw the report. I haven't tried reverting and then reproducing it.

Unrelated: if you use IRC/Matrix, please consider joining the #etesync channel. Would be great to have you there. Links: https://www.etesync.com/community-chat/

PR #150, opened 2020-07-21
Birthday Calendar

### LINKED ISSUE #77: Birthday calendar: add a generated calendar for address book birthdays
The calendar shouldn't be synced or even visible in the journals list. It should just be generated locally from the address book birthdays and shown in the etesync web calendar.

@Tal-Leibman maybe you'd like to take a look?

ramzan:
Fixes #77.

### DIFF
--- a/src/Pim/PimMain.tsx
+++ b/src/Pim/PimMain.tsx
@@ -67,8 +67,10 @@ class PimMain extends React.PureComponent<PropsType> {
     // FIXME:Hack
     const itemUid = `${(event as any).journalUid}|${encodeURIComponent(event.uid)}`;
 
-    this.props.history!.push(
-      routeResolver.getRoute('pim.events._id', { itemUid }));
+    if (event.uid) {
+      this.props.history!.push(
+        routeResolver.getRoute('pim.events._id', { itemUid }));
+    }
   }
 
   public taskClicked(event: ICAL.Event) {
--- a/src/Pim/PimMain.tsx
+++ b/src/Pim/PimMain.tsx
@@ -64,10 +64,10 @@ class PimMain extends React.PureComponent<PropsType> {
   }
 
   public eventClicked(event: ICAL.Event) {
-    // FIXME:Hack
-    const itemUid = `${(event as any).journalUid}|${encodeURIComponent(event.uid)}`;
-
     if (event.uid) {
+      // FIXME:Hack
+      const itemUid = `${(event as any).journalUid}|${encodeURIComponent(event.uid)}`;
+
       this.props.history!.push(
         routeResolver.getRoute('pim.events._id', { itemUid }));
     }
--- a/src/Pim/index.tsx
+++ b/src/Pim/index.tsx
@@ -19,7 +19,7 @@ import { createSelector } from 'reselect';
 
 import { History } from 'history';
 
-import { PimType, ContactType, EventType, TaskType } from '../pim-types';
+import { PimType, ContactType, EventType, TaskType, parseString } from '../pim-types';
 
 import Container from '../widgets/Container';
 
@@ -395,6 +395,25 @@ class Pim extends React.PureComponent {
     const { collectionsAddressBook, collectionsCalendar, collectionsTaskList, addressBookItems, calendarItems, taskListItems } = itemsSelector(this.props);
     const [contacts, events, tasks] = itemValuesSelector(this.props);
 
+    const birthdays = contacts.filter((c) => c.bday).map((c) => {
+      let { year, month, day } = c.bday._time;
+      year = year || '2020';
+      if (month < 10) {
+        month = '0' + month;
+      }
+      if (day < 10) {
+        day = '0' + day;
+      }
+
+      const event = 
+        'BEGIN:VEVENT\n' +
+        `SUMMARY:${c.fn}'s Birthday\n` +
+        `DTSTART;VALUE=DATE:${year}${month}${day}\n` +
+        'RRULE:FREQ=YEARLY\n' +
+        'END:VEVENT';
+      return new EventType(parseString(event));
+    });
+
     return (
       <Switch>
         <Route
@@ -403,7 +422,7 @@ class Pim extends React.PureComponent {
           render={({ history }) => (
             <PimMain
               contacts={contacts}
-              events={events}
+              events={events.concat(birthdays)}
               tasks={tasks}
               history={history}
               onItemSave={this.onItemSave}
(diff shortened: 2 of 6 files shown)


src/Pim/PimMain.tsx
```diff
@@ -67,8 +67,10 @@ class PimMain extends React.PureComponent<PropsType> {
     // FIXME:Hack
     const itemUid = `${(event as any).journalUid}|${encodeURIComponent(event.uid)}`;
 
-    this.props.history!.push(
-      routeResolver.getRoute('pim.events._id', { itemUid }));
+    if (event.uid) {
```
tasn, line 70: I think it should be clickable, just the "edit" button shouldn't be clickable in the event page itself.

src/Pim/index.tsx
```diff
@@ -395,6 +395,25 @@ class Pim extends React.PureComponent {
     const { collectionsAddressBook, collectionsCalendar, collectionsTaskList, addressBookItems, calendarItems, taskListItems } = itemsSelector(this.props);
     const [contacts, events, tasks] = itemValuesSelector(this.props);
 
+    const birthdays = contacts.filter((c) => c.bday).map((c) => {
+      let { year, month, day } = c.bday._time;
+      year = year || '2020';
```
tasn, line 400: Having the default year as 2020 feels wrong. It should probably just be a recurring event starting from the person's date of birth and if not, start from 1900.

tasn, line 400: Another thing, does this actually work with birthdays without the year set? It's great if it does. What's the type of `bday`? (Actualy type, not typescript type). I remember that it could change and thus `_time` wouldn't always be available. If it doesn't, then you can just change `bday` below to return the correct type. (`getFirstPropertyValue<ICAL.Time>(...)`) and then work with it. Though I remember that we had to hack around it in the iOS client. Take a look at what we did there: https://github.com/etesync/ios/blob/master/src/sync/helpers.ts#L417

src/Pim/index.tsx
```diff
@@ -395,6 +395,25 @@ class Pim extends React.PureComponent {
...
+      if (month < 10) {
+        month = '0' + month;
+      }
+      if (day < 10) {
+        day = '0' + day;
```
tasn, line 405: Even though the typing of `_time` is any, which is why you can do such things, it's not something we do (changing of type). Additionally, you can simplify all of this to:
```
const bdayTime = c.bday._time;
const year = bdayTime.year ?? '2020';
const month = bdayTime.month.toString().padStart(2, '0');
const day = bdayTime.day.toString().padStart(2, '0');
```

src/Pim/index.tsx
```diff
@@ -395,6 +395,25 @@ class Pim extends React.PureComponent {
...
+      }
+
+      const event = 
+        'BEGIN:VEVENT\n' +
+        `SUMMARY:${c.fn}'s Birthday\n` +
+        `DTSTART;VALUE=DATE:${year}${month}${day}\n` +
+        'RRULE:FREQ=YEARLY\n' +
+        'END:VEVENT';
+      return new EventType(parseString(event));
+    });
```
tasn, line 415: Creating events and parsing them like this is probably not the best way of doing it.
Please just construct the event here, it's really just a matter of doing:
```
const event = new EventType();
event.uid = uuid.v4();
event.summary = `${c.fn}`s Birthday`;
event.startDate = timeNativeToVobject(date);
const rrule: ICAL.RecurData = {
    freq: 'YEARLY',
    interval: 1,
  };
event.rrule = new ICAL.Recur(rrule);
return event;
```
Where `timeNativeToVobject` is https://github.com/etesync/ios/blob/master/src/sync/helpers.ts#L192

src/Pim/index.tsx
tasn: Having given it more thought, I'm pretty sure it's indeed the case, and that you should use `parsDate` from the link. I remember opening a ticket about it on ICAL.js's bug tracker.

src/Pim/PimMain.tsx
ramzan: How should it handle the journalUid being undefined?

tasn: That's actually another thing I wanted to raise, but thought maybe I would do it in a future PR, though I guess it's already an issue!
I think the birthday calendar should be added earlier in the chain. You are now adding it as part of the events, but I think it should be added as a fake journal altogether. Adding it there is not that different, you just need to add a new calendar journal when decrypting the data (SyncGate.tsx? not sure), make it read-only, and set the color and name as you want it. That's it. :)

src/Pim/index.tsx
```diff
@@ -395,25 +398,30 @@ class Pim extends React.PureComponent {
...
-        `SUMMARY:${c.fn}'s Birthday\n` +
-        `DTSTART;VALUE=DATE:${year}${month}${day}\n` +
-        'RRULE:FREQ=YEARLY\n' +
-        'END:VEVENT';
-      return new EventType(parseString(event));
-    });
-
+    let birthdays: EventType[] = [];
+
+    if (store.getState().settings.showBirthdayCalendar) {
```
tasn, line 403: This is not the correct way to do it, because it won't update when the store updates.
It's actually very annoying to use Redux without react hooks though, so maybe if you can't easily fix it to get it from the store, let's leave it like this and add a "FIXME" comment, so when we change it to hooks (I can do it after this PR is merged) it will be fixed.

src/store/reducers.ts
```diff
@@ -273,6 +273,7 @@ export const errorsReducer = handleActions(
 export interface SettingsType {
   locale: string;
   darkMode?: boolean;
+  showBirthdayCalendar?: boolean;
```
tasn, line 276: At the moment it won't show for people by default. You could implement a migration to set a default value for it, but I guess an easier solution would be to flip this field.
So `showBirthdayCalendar` -> `hideBirthdayCalendar` and have the default be `false` which is great because `undefined` is also falsey.

src/Pim/index.tsx
```diff
@@ -395,6 +398,30 @@ class Pim extends React.PureComponent {
...
+          event.summary = `${c.fn}'s Birthday`;
+          event.startDate = timeNativeToVobject(date, true);
+          const rrule: ICAL.RecurData = {
+            freq: 'YEARLY',
+            interval: 1,
+          };
+          event.rrule = new ICAL.Recur(rrule);
+        }
+        return event;
+      }).filter((ev) => ev.uid !== undefined);
```
tasn, line 422: Maybe I'm blind, but I can't see you setting the event uid at all.
Additionally, you probably want to change this filter to `filter(isDefined)` using this function: https://github.com/etesync/ios/blob/master/src/helpers.tsx#L110
and just return undefined as the first step once you realise this event can't be parsed. Without even creating the event in the first place.

tasn:
Hey, what's up? Just following up on this, as it would be great to have it!

PR #159, opened 2020-08-19
Birthday calendar 2

### LINKED ISSUE #77: Birthday calendar: add a generated calendar for address book birthdays

ramzan:
Hello, I am back! My last PR was a bit messy so I decided to make a new one.

This time I created the fake journal as you suggested.

The hideBirthday setting gets passed to Pim as a prop so it re-renders now.

The birthday calendar events are now clickable but the edit/duplicate buttons are disabled.

Fixes #77

### DIFF
--- a/src/SyncGate.tsx
+++ b/src/SyncGate.tsx
@@ -26,6 +26,9 @@ import { CURRENT_VERSION } from "etesync";
 
 import { store, JournalsData, EntriesData, StoreState, CredentialsData, UserInfoData } from "./store";
 import { addJournal, fetchAll, fetchEntries, fetchUserInfo, createUserInfo } from "./store/actions";
+import { syncEntriesToItemMap } from "./journal-processors";
+import { ContactType } from "./pim-types";
+import { parseDate } from "./helpers";
 
 export interface SyncInfoJournal {
   journal: EteSync.Journal;
@@ -45,6 +48,7 @@ interface SelectorProps {
   journals: JournalsData;
   entries: EntriesData;
   userInfo: UserInfoData;
+  hideBirthdays?: boolean;
 }
 
 const syncInfoSelector = createSelector(
@@ -52,7 +56,8 @@ const syncInfoSelector = createSelector(
   (props: SelectorProps) => props.journals!,
   (props: SelectorProps) => props.entries,
   (props: SelectorProps) => props.userInfo,
-  (etesync, journals, entries, userInfo) => {
+  (props: SelectorProps) => props.hideBirthdays,
+  (etesync, journals, entries, userInfo, hideBirthdays) => {
     const derived = etesync.encryptionKey;
     const userInfoCryptoManager = userInfo.getCryptoManager(etesync.encryptionKey);
     try {
@@ -65,7 +70,9 @@ const syncInfoSelector = createSelector(
       }
     }
 
-    return journals.reduce(
+    let bdayEntries = List<EteSync.SyncEntry>();
+
+    let syncInfo = journals.reduce(
       (ret, journal) => {
         const journalEntries = entries.get(journal.uid);
         let prevUid: string | null = null;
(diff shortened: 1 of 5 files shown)


src/SyncGate.tsx
```diff
@@ -95,6 +105,46 @@ const syncInfoSelector = createSelector(
...
+        return;
+      }
+      const year = bdayTime.year ?? 1900;
+      const month = (bdayTime.month + 1).toString().padStart(2, "0");
+      const day = bdayTime.day.toString().padStart(2, "0");
+
+      const event =
+        "BEGIN:VCALENDAR\nBEGIN:VEVENT\n" +
+        `SUMMARY:${c.fn}'s Birthday\n` +
+        `UID:${uuid.v4()}\n` +
```
tasn, line 133: Just a thought (maybe a bad idea), but wouldn't it be better to use the contact's uid here instead of a randomly generated one? This will make it persistent across refreshes.

src/Pim/index.tsx
```diff
@@ -247,6 +247,7 @@ const CollectionRoutes = withStyles(styles)(withRouter(
               if (this.itemUndefined(itemUid)) {
                 return PageNotFound();
               }
+              const isBirthday = itemUid.match(/^birthdays\|*/) !== null;
```
tasn, line 250: I think this is a bit hacky, but more than that, it's unnecessary.
It can just be done in SyncGate. Just make the selector also depend on this setting and only generate it if the setting is set. If you make the journal read-only the no-editing will already be handled for you (if it doesn't that's a bug in etesync-web that can be fixed).
This will make it so you don't even need to touch this file.

tasn (changes requested):
Great job! I love this change, it's super concise and looks great!
I made some comments on how this can be simplified further. Please let me know if you have any questions.

src/SyncGate.tsx
ramzan: I rearranged the code a bit to allow the uid to be a combination of the journal uid and contact uid. This should allow persistence and avoid collisions.

src/Pim/index.tsx
ramzan: Done, but the editing buttons are still enabled.

src/SyncGate.tsx
tasn: Great!

src/Pim/index.tsx
tasn: I just checked, and the code indeed doesn't deal with readOnly. I'll fix that in a moment.

tasn:
I squashed some commits together, but otherwise all was great! Good job, so nice to finally have this!

tasn:
I just found a bug! I'm not sure if I'm correct because I can't test it at the moment, but what happens if you delete a contact, is it still shown in the birthday calendar? I believe the answer is yes, which is not good. The problem is that this function works on the journal entries rather than the items so old entries are also processed.

tasn:
Another thought, maybe the right thing to do is a combination of what you did before (having it in `Pim/index.tsx`) and what we did now, by monkey patching syncInfo. What I mean is, do the creation of the birthday journal in `Pim/index.tsx` and set it on syncInfo like you just did directly from pim/index. This will mean you will only have the actual items to work with so this bug won't happen, but it'll also be set on syncinfo so it'll be generic. A bit hacky, but the best of both worlds.
I'll happily do it if you don't have the time, please let me know.

ramzan:
I just tested it out and the event does disappear when the contact is deleted.

tasn:
Ah oops, yeah, you are calling `syncEntriesToItemMap` so it should be fine. Sorry, was tired yesterday. :)
Btw, I fixed the readOnly bug! Thanks a lot, this is such an often request feature!

Quantenzitrone:
where do i access this? i can't find it on pim.etesync.com

Issue #138, opened 2020-07-03
Can't Change Calendar After Savings


kevin2themystic:
Please add the ability to allow users to:
-  change the calendar of a certain event after it has been saved
- set a default calendar

tasn:
Hey,

Yeah, it would be great! Sorry for not replying earlier...

Until this is implemented, you can duplicate the event, change the calendar, and then delete the old one. Not ideal, but this is essentially what will happen behind the scenes once this is implemented.

PR #162, opened 2020-08-30
Allow changing calendar when editing event

### LINKED ISSUE #141: Feature request: moving events between calendars
See [moving events between calendars](https://github.com/etesync/server/issues/49)

ramzan:
Fixes #141 and partially #138, which also requests a default calendar option.

### DIFF
--- a/src/components/EventEdit.tsx
+++ b/src/components/EventEdit.tsx
@@ -219,7 +219,9 @@ class EventEdit extends React.PureComponent<PropsType> {
       new EventType()
       ;
 
-    event.uid = this.state.uid;
+    const collectionChanged = !this.props.duplicate && this.props.initialCollection && this.state.journalUid !== this.props.initialCollection;
+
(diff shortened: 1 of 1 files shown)


tasn:
I'm on the fence about this change. I wonder if it's better to have a "move to another collection" option that will be available for contacts, calendars and tasks instead of transparently deleting and adding to another collection. The most obvious way where it's matter is the change history which is lost - so users need to at least be aware of this.

tasn:
Btw, I don't know if you saw the `etebase` branch, but that's the etesync 2.0 branch that cleaned up a lot of the code! I'll probably have it ready for testing later this week so maybe consider starting developing there from then on.

ramzan:
Yeah that seems like a better idea. I'll close this PR and check out `etebase`.

tasn:
Just FYI: I haven't ported over all of your PRs yet.

PR #167, opened 2020-10-08
Fix localization for DateTimePicker

niconfus:
Use date format based on the locale setting for adding/editing a calendar event instead of hardcoding it.

Tested by making sure the observed format was "DD/MM/YYYY" when the locale is English (United Kingdom) and "MM/DD/YYYY" when the locale is English (United States).


tasn:
Oops, nice catch!

Issue #136, opened 2020-06-30
Edit contact groups with WebApp?


koreywithak:
I know the WebApp is in Beta, but I was wondering if there were any plans to allow editing of contact groups? It would be nice to be able to edit them online. It does not seem possible currently. Otherwise I think I have everything I need and I appreciate the E2E calendar/contact capability! Thanks.

piu-logistics:
I'd like to fully adopt Ete, working with v2, generally great idea, nice one.  

We use Google Contacts a lot and need the Group features to be accessible via the Ete web UI.  

We've moved away from Thunderbird and don't really want to return to that or Evolution clients.  A good online Address Book featuring Group edit / viewing functions would remove the key obstacle to fully using Ete (and dropping Google).  Please prioritise this update as a NOW issue, more people will adopt EteSync with a complete UI.

piu-logistics:
Let me expand on our situation:

We have several G accounts, each with their own G Contacts.

1) One central contact list would make it much more efficient and effective to administer, but access must be online (not bound to a specific pc or to handheld devices).
2) Centralising needs stability and security as a matter of basic risk management.
3) Google has a systemic problem with contact syncing, i have read others also complaining about it.  To the degree that there are apps that claim/aim to "Fix Google Contact Syncing". Those have not worked for us. Maybe due to the relatively large about of contacts, 5k+.  Why shouldn't it work with 100k+, we'd like that capacity.

What we need:

4) When you open Address Book you should see two columns, Groups and Contacts.
5) The first selection under groups would be "All" and naturally that shows all contacts in the Contacts section.  Then below all you have the rest of the groups in alphabetical order.
6) It should be ease to add a contact to any existing group or to create a new group; from both the single contact view and the groups/contacts view.
7) We need to be able to import and export groups of contacts, retaining their Group labels.

That's all that we need.

To win attention for Ete for its unique features then...

8) Make it so that i can share access to specific groups with named individuals; Often we have teams working on different matters and they need access to specific contact lists.  Mostly this is when the lists are being curated/formed/built.  Such access/users should not be able to export the group listing or see other lists.  I like that Ete journals change, if they mess stuff up it can be fixed!
9) Find a way to sync up contacts, cal and tasks with Protonmail

tasn:
I'm deploying a change with viewing groups.

One comment:
> Find a way to sync up contacts, cal and tasks with Protonmail (the new Beta looks good)

[...]

PR #176, opened 2020-10-29
Contacts: Add basic group editing functionality

ramzan:
The basic functionality requested in #136 is here: creating/editing/deleting groups and adding/removing members.

 The UI still needs some work, but this should work for now.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -41,7 +41,10 @@ export const routeResolver = new RouteResolver({
     contacts: {
       _id: {
         _base: ":itemUid",
-        edit: "edit",
+        edit: {
+          contact: "contact",
+          group: "group",
+        },
         log: "log",
       },
       new: {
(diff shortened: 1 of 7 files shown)


tasn:
Oh wow, you are quick, good job!
I like what you did with adding `contact` and `group` to the URL, seems like the good approach. Why didn't you do it for edit too? How are groups edited then? Just directly from the contacts field?

My main comment is: why have them both in ContactEdit? Why not have ContactEdit and GroupEdit as separate views (You can even redirect between them if you get the wrong type)? They don't seem to be sharing much code anyway, and it's similar to how we split the rest of the views.

Looking forward to hearing your thoughts..

ramzan:
I agree with all your comments. The reason I stuffed everything into ContactEdit is that I was preoccupied with figuring it out/making it work and forgot about good design lol. I'll split it into separate views, no problem.

tasn:
Hehe, looking forward to the next version! :)

ramzan:
Should be good now

tasn:
Sorry it took me a few days to review, I wanted to give it a proper look, though everything is great, gj!

PR #178, opened 2020-11-05
ContactEdit: fix contacts getting added to groups multiple times

ramzan:
Mixed up an array with a generic Object at one point, leading to line 365 always being true.


tasn:
Thanks!

PR #181, opened 2020-11-09
Add Docker Packaging

BeatLink:


tasn:
Hey, thanks for the PR!
This is something that has been brought up before and I'm still on the fence about. I don't think having a docker image that just serves static files really adds a lot of benefit to users compared to the extra burden of having to maintain it. You could argue that it's good to have an image to build the web app, and I see benefit in that. Though the extra nginx part feels a bit too lightweight to me.
I can see how this essentially becomes a complex image (to support e.g. SSL) with a lot of configurations for running it, all to server a bunch of static files.

Would love to hear your thoughts...

BeatLink:
Well the way I see it, its just one other packaging scheme thats available to end users. The more the merrier I think! The more packaging methods available, the more convenient it is. Right now, anyone can simply grab this image and spin it up without needing to mess with node or yarn or such

tasn:
We already build the web app as a tar.gz you can just download at https://pim.etesync.com/etesync-web.tgz so you don't really need to mess with node or yarn, unless you want to build for yourself. Though then, I don't think it's such a big ask...

I generally agree with the more the merrier, and I encourage you to build this docker image and release it, though I'm not sure if I want to take over the burden of longer term maintenance of it. These things add up. Take a look at etesync-dav for example. It's the more the merrier, but I now need to package it for AUR, Docker, PyPi, and bundle for Windows, mac and Linux. These things add up...

tasn:
Decided against this?

BeatLink:
Oh, well not quite. I do have a dockerfile that would build the image from a node container then spin up an nginx container and run it from there, but i am still trying to iron out the kinks in terms of specifying the server URL

BeatLink:
I'll recreate the dockerfile

PR #189, opened 2020-12-04
Add Docker Build

BeatLink:
This should reopen the pull request to add the docker build. 

I understand that you are reluctant to take on the packaging for docker as the web client is nothing more than a bundle of static files. But this docker is very simple, All it does is pull the latest version of the web client from git and build it in a node docker container, then copies the built files from that build container into an official nginx container. It will require little to no maintenace. 

In addition, seeing as there are docker images for both the server and the dav bridge, it would make logical sense to have one as well for the web client. This would fulfill that role. 

I am willing to take on maintenance for any bug reports or feature requests for the docker package. All you have to do is assign the relevant issues to me. 

I just really wish to see an official web client image that users can use and trust. The dockerfile would be in the codebase for all to see and verify for themselves and it would simplify installation and self hosting.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,17 @@
+# Builds the Web Client from the latest git repo
+
+# Build Stage =====================================================================================
+FROM node:alpine as build
+ENV DEFAULT_ETEBASE_SERVER "http://example.com"
+RUN export REACT_APP_DEFAULT_API_PATH=${DEFAULT_ETEBASE_SERVER} && \
+    git clone https://github.com/etesync/etesync-web.git etesync-web && \
+    cd etesync-web && \
+    yarn && \
+    yarn build
+
+# Run Stage =======================================================================================
+FROM nginx:alpine
+COPY --from=build /etesync-web/build /usr/share/nginx/html
+COPY ./nginx.conf /etc/nginx/conf.d/default.conf
+EXPOSE 80
+EOF
--- a/docker/nginx.conf
+++ b/docker/nginx.conf
@@ -1,0 +1,17 @@
+server {
+
+    listen 80 default_server;
+    listen [::]:80 default_server;
+    charset utf-8;
+    client_max_body_size 75M;     # max upload size
+    location /        {
+      root /usr/share/nginx/html;
+      try_files $uri $uri/ /index.html =404;
+      autoindex on;
+    }
+    error_page   500 502 503 504  /50x.html;
+    location = /50x.html {
+        root   /usr/share/nginx/html;
+    }
+}
+EOF


BeatLink:
Here's a link to my test dockerhub page, though ideally, it would be hoped to be migrated to the official etesync dockerhub account. 

https://hub.docker.com/r/beatlink/etesync-web-docker

tasn:
> In addition, seeing as there are docker images for both the server and the dav bridge, it would make logical sense to have one as well for the web client. This would fulfill that role.

I don't think it's the same thing. The server is actually complex, comes with deps, configurations and etc. The DAV bridge stores some data internally and comes with deps too. To be honest, the docker image was created before we had pre-built binaries. Not sure if we would have done it if we already had binaries.

Web, on the other hand, is really just a bunch of files that we already package and sign pre-built, as well as offer the option to build yourself.

> I just really wish to see an official web client image that users can use and trust. The dockerfile would be in the codebase for all to see and verify for themselves and it would simplify installation and self hosting.

This is a very good point, because even though I don't agree with the need, you are not the only one that wants a docker container and having an official one will therefore be better. The question is then: what about etesync-notes? That should probably also be in the same docker container, no? How should it work? Different ports? I'd rather not have different paths because I don't want to commit to supporting that in the images.
Additionally, does this mean this docker image sits here or elsewhere?

Could you remind me on Wednesday if I don't comment here until then? Thanks!

BeatLink:
Well we could do like etesync dav and store the dockerfile for each application in the source code repo itself. And each dockerfile would build each specific app rather than bundling the different applications together in a single dockerfile. That way we follow the docker philosopy of one service, one container.

tasn:
Yeah, but this is not really a service, it's just files. :P 

It would just be a pity if people would need to run two containers, and two local nginx instances just to serve a few static files.

BeatLink:
Well the alternative would be to build an omnibus package containing both the server and the web client and rolling everything into one

BeatLink:
Which is what I was asking for in the first place hahahahaha :joy:

Ezwen:
Hi there! Just wanted to mention that I was indeed surprised to not find any official Docker image for etesync-web, and that imho it would be great if there was one. I personally user docker for 100% of the services I host, so that would make things really easier in my case :).

> It would just be a pity if people would need to run two containers, and two local nginx instances just to serve a few static files.

It's true that with docker we tend to spawn lots of nginxes in different containers for various web services − even static ones such as this one − which can be perceived as a waste. But nowadays this is rather well accepted since nginx only has a tiny footprint, and since it makes managing web services easier, when each lives its own process, each with its own nginx or apache version, each with its own configuration.

tasn:
OK, I'm convinced. Just a few comments about the PR then:

1. Can we make it copy the current directory and build it from that instead of from git? 
2. Can we make it copy package.json and yarn.lock before the rest and then run `yarn install` so that this part is cached?

Thanks!

tiritibambix:
Hello. Does that mean there is no way to host etesync-web with docker ?

tasn:
@tiritibambix, it's just a static website, you can just build it and host it. You don't really need docker for this (though you can).

BadCo-NZ:
@BeatLink any update on your progress or response to @tasn comments above?

I came here looking for a docker image of the web app after replacing nextcloud with etesync, looking forward to the containerisation! 😁

BeatLink:
Unfortunately, I'm not currently using Etesync right now

michael2to3:
I agree, I would like to just insert the docker compose file into deploy and work with the program

PR #226, opened 2021-09-04
Improve PWA manifest

2br-2b:
This started when installing the PWA directed me to index.html, which gave me a 404 error. I decided to fix that, and while I was at it, I figured I might as well add a few more fields to the manifest!

All of the specs added were added as documented in https://developer.mozilla.org/en-US/docs/Web/Manifest.

The only improvement I can see to add would be [screenshots](https://developer.mozilla.org/en-US/docs/Web/Manifest/screenshots), but as my calendar, tasks, etc. are filled with my personal info, I'd prefer not to upload my own screenshots :-). Is there any way you could generate some screenshots on a dummy account and host them on your website, Github, or IPFS so that they can be pulled in by the manifest?

### DIFF
--- a/public/manifest.json
+++ b/public/manifest.json
@@ -8,7 +8,7 @@
       "type": "image/png"
     }
   ],
-  "start_url": "./index.html",
+  "start_url": "https://pim.etesync.com/",
   "display": "standalone",
   "theme_color": "#ffc107",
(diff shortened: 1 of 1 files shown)


2br-2b:
Actually, see #227 for some other slight improvements to the manifest.

PR #234, opened 2021-12-13
fix event descriptions to respect whitespace

jzacsh:
I haven't actually stood up a local server yet to test this, but thought I'd hit send anyway. LMK if looks good and then I'll figure out how to test locally (just... in the wildly-off-chance somehow I caused a react bug or something)

### DIFF
--- a/src/Calendars/Event.tsx
+++ b/src/Calendars/Event.tsx
@@ -35,7 +35,7 @@ class Event extends React.PureComponent {
           <div><u>{this.props.item.location}</u></div>
         </PimItemHeader>
         <div style={style.content}>
-          <p style={{ wordWrap: "break-word" }}>{this.props.item.description}</p>
+          <p style={{ wordWrap: "break-word", whiteSpace: "pre-wrap" }}>{this.props.item.description}</p>
           {(this.props.item.attendees.length > 0) && (
             <div>Attendees: {this.props.item.attendees.map((x) => (x.getFirstValue())).join(", ")}</div>)}
         </div>


tasn:
This is a potential to look very bad (though I guess also OK if we assume everything is pre-formatted). What was the exact issue you were facing?

jzacsh:
> This is a potential to look very bad (though I guess also OK if we assume everything is pre-formatted).

Oh interesting. I'm having trouble picturing an example where this would cause issues. Are you thinking like... some auto-generated content that inserted a ton of linebreaks (and now the etesync UI would take up multiple scrolling pages)? Or do you have an example in mind of human-written events that would have problems?

> What was the exact issue you were facing?

I guess it could just be me. Most of my events actually only look bad in the etesync UI where the text is assumed to be whitespace insignificant. That is: all calendar UIs I've used I always add line breaks and other simple things to help myself read my own descriptions, knowing that there's no rich formatting in calendars. (well Google calendar has some limited richtext ability now, but it's always finicky for me so I've never bothered relying on it).

**tl;dr** many of my events' descriptions are illegible unless whitespace is significant (and etesync is the first time I've encountered this).


----


If this can indeed cause problems maybe this idea should be punted to a feature request where users can toggle this as a preference? Though at that point it's probably not worth it(?) since this is just the webui which I gather isn't really intended to be feature rich.

jzacsh:
(marking this closed, because I don't think this is worth breaking anyone else just to fix me; but feel free to reopen)

PR #241, opened 2022-01-27
Improve Layout of ContactEdit

b-erhart:
I thought the current ContactEdit screen looks a bit messy, since the alignments and margins were a bit off in some places. So I tried to improve the situation a little.

This is a short summary of the changes I made:
- Use `Grid`-based layout
- Change `TextField` `placeholder` attributes to `label` so the name is still visible after input
- Switch `TextField` `variant`s from `standard` to `outlined` for a more modern look and more even looking spacing
- Add icons for better orientation
- Add button to show/hide name fields that are rarely used. Fields that have content are never hidden

Obviously, design stuff is a pretty subjective thing. So I'm not sure if you like this kind of layout or if this is the design direction you want to go in. This is simply a suggestion and I am happy to make adjustments.

Here is a small comparison:
Before:
ContactEdit_before
After:
ContactEdit_after_new

### DIFF
--- a/src/Contacts/ContactEdit.tsx
+++ b/src/Contacts/ContactEdit.tsx
@@ -95,7 +95,7 @@ interface ValueTypeComponentProps {
 
   types: Array<{ type: string }>;
   name: string;
-  placeholder: string;
+  label: string;
   value: ValueType;
   onClearRequest: (name: string) => void;
   onChange: (name: string, type: string, value: string) => void;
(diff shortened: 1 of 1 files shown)


b-erhart:
Also, I tried to follow the general style of the existing code. But since this is my first PR here, I'm not sure if I managed to do that everywhere. Please tell me if something doesn't fit and I will change it.

tasn:
Oh, why did you close it? Sorry I missed it until now. Want me to take a look?

PR #259, opened 2023-05-14
feat: add auto dark mode

oleobal:
Add detection for theme preferences given by the browser.

This works by replacing the "dark mode" setting with a new selector that defaults to "auto" (losing old settings, sorry).

You can try this out by opening the dev console and using the dedicated button (Inspector > sun/moon -shaped in Firefox, Elements > paint brush -shaped in Chromium).

I tried to match the project style but I'm not too good at frontend, I hope this is OK!

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -150,21 +150,30 @@ export default function App() {
   const fetchCount = useSelector((state: store.StoreState) => state.fetchCount);
   const errors = useSelector((state: store.StoreState) => state.errors);
 
-  function shouldBeDark(userSelection: string): boolean {
+  function shouldBeDark(userSelection: string, browserPreference: boolean): boolean {
     if (userSelection === "auto") {
-      return window.matchMedia && window.matchMedia("(prefers-color-scheme: dark)").matches;
+      return browserPreference;
     } else if (userSelection === "dark") {
       return true;
     }
     return false;
   }
-  const [darkMode, setDarkMode] = React.useState(() => shouldBeDark(darkModeUserSelection));
+
+  const [darkModeBrowserPreference, setDarkModeBrowserPreference] = React.useState(Boolean(window.matchMedia?.("(prefers-color-scheme: dark)").matches));
+  const handleBrowserDarkModePreferenceChange = React.useCallback((e) => {
+    setDarkModeBrowserPreference(e.matches);
+  }, []);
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -146,10 +146,26 @@ export default function App() {
...
+    } else if (userSelection === "dark") {
+      return true;
+    }
+    return false;
+  }
+  const [darkMode, setDarkMode] = React.useState(() => shouldBeDark(darkModeUserSelection));
+  React.useEffect(() => {
+    setDarkMode(shouldBeDark(darkModeUserSelection));
+  }, [darkModeUserSelection]);
+  window.matchMedia && window.matchMedia("(prefers-color-scheme: dark)").addEventListener("change", () => {
```
tasn, line 165: This listener should be added in its own `useEffect` and removed when the useEffect destructor is called.
It should also probably just set a local state variable `darkModeBrowserPrefers` or something which defaults to the current value.

src/App.tsx
```diff
@@ -146,10 +146,26 @@ export default function App() {
...
   const dispatch = useDispatch();
   const etebase = useCredentials();
-  const darkMode = useSelector((state: store.StoreState) => state.settings.darkMode);
+  const darkModeUserSelection = useSelector((state: store.StoreState) => state.settings.darkModeUserSelection);
   const fetchCount = useSelector((state: store.StoreState) => state.fetchCount);
   const errors = useSelector((state: store.StoreState) => state.errors);
 
+  function shouldBeDark(userSelection: string): boolean {
+    if (userSelection === "auto") {
+      return window.matchMedia && window.matchMedia("(prefers-color-scheme: dark)").matches;
```
tasn, line 155:
```suggestion
      return window?.matchMedia("(prefers-color-scheme: dark)").matches;
```
I think that's the short syntax, though maybe it's `.?` please double check.

tasn:
Thanks for the contribution! I made some suggestions.

src/App.tsx
oleobal: Thanks for the pointer, I didn't about this so I looked things up, I hope it's better now. It certainly looks cleaner in my opinion.

tasn: Yup, looks great!

tasn:
Deployed, thanks!

